PR #7, opened 2023-06-24
develops must call analysis

Kevin-TD:
-must call pass currently sloppily pasted onto called methods pass; updates to code structure soon


src/Analysis.cpp
```diff
@@ -25,8 +25,9 @@
 // or to run all tests: sh ../run_all.sh 
 // note for run all tests is that if you add more tests, you'll have to modify run_all.sh to include that test number
 
-// updated: automated testing, aliasing, checking if inputs have changed 
-// todo: must call pass; what is done for branching if in one branch malloc is called and another it is not? just need it explained
+// TODO: how to handle unknown functions 
+// TODO: combine analysis such that MC & CM run at same time OR make them two separate things
```
kelloggm, line 29: Do the second one: keep them separate, like we discussed in our call today.

src/Analysis.cpp
```diff
@@ -42,6 +43,16 @@ typedef std::map<std::string, std::map<std::string, CM>> CalledMethods; // left
 typedef std::map<std::string, std::string> AliasMap;
 typedef std::map<std::string, std::string> VarBranchMap;
 
+// what do we do when unknown function is called? 
+// do we add realloc to mc set?
```
kelloggm, line 47: It depends on what you mean by "unknown function" and "realloc". Can you make this a bit more specific so I can be sure I'm answering the right question?

src/Analysis.cpp
```diff
@@ -802,10 +813,381 @@ void CalledMethodsAnalysis::doAnalysis(Function &F, PointerAnalysis *PA, std::st
...
 
   errs() << "\n\nRUNNING TESTS - ALLOWED_REDEFINE = " << ALLOW_REDEFINE << " TEST NAME - " << testName << "\n\n";
-  std::exit(runTests(ExpectedResult, PostCalledMethods));
+  // * Tests temporarily disabled; developing must call first and then changing structure 
+  // std::exit(runTests(ExpectedResult, PostCalledMethods));
+
+}
+
+// must call 
+void transfer(Instruction* I, SetVector<Instruction *>& workSet, std::map<std::string, MC>& mustCallSet, AliasMap& aliasedVars) {
```
kelloggm, line 822: I think a lot of this function's code is shared with `transfer` in the called-methods analysis. Would it be possible to abstract out a "generic" transfer function that works for both of them, and have it call methods for the details that need to differ? That would be better software engineering than just copying code between the two.

kelloggm (changes requested):
Before I review this further, I want the code to be restructured:
* the two analyses should be split up into different classes that both implement the same interface (which you'll need to do with an abstract class in C++, IIRC)
* code that both analyses need should exist in only one place, and should be called by both. The correct place for such code is the shared abstract class.

src/Analysis.cpp
Kevin-TD: Basically, if we first make a call to malloc then call something like realloc, do we add realloc to the must call set or no?

kelloggm: Calling `realloc` doesn't change the need to call `free`, does it? So I'm not sure that it matters (although it does create a new obligation: you need to erase the called methods set at calls to `realloc` or any other method that might re-allocate the pointer)

include/PassType.h
```diff
@@ -0,0 +1,51 @@
...
+private: 
+    AliasMap aliasedVars; 
+    std::string testName; 
+    CFG* cfg; 
+    MappedMethods expectedResult; 
+    
+    virtual void analyzeCFG(CFG* cfg, MappedMethods& PreMappedMethods, MappedMethods& PostMappedMethods, std::string priorBranch) = 0; 
+    virtual void transfer(Instruction* I,  SetVector<Instruction *> workSet, std::map<std::string, MethodsSetHolder>& inputMethodsSet) = 0;
+public:    
+    void setFunctions(std::map<std::string, bool> safeFunctions, std::map<std::string, bool> unsafeFunctions, std::map<std::string, bool> reallocFunctions, std::map<std::string, std::string> memoryFunctions);
```
kelloggm, line 31: Why do these need to be maps from `string->bool`? I would think just keeping the set of names (as strings) would suffice, without the need for the bool. If the bool is really needed, what it represents needs to be explained in a comment.

include/PassType.h
```diff
@@ -0,0 +1,51 @@
...
+    void buildExpectedResult(std::string testName); 
+
+    std::map<std::string, bool> safeFunctions;
+    std::map<std::string, bool> unsafeFunctions; 
+    std::map<std::string, bool> reallocFunctions;
+    std::map<std::string, std::string> memoryFunctions;
+    std::string passName; 
+
+   
+    bool runTests(MappedMethods receivedResult);
```
kelloggm, line 41: Why is running the tests something that a pass can do? I would expect `runOnCFG` (or something similar) as a method within the pass, and then have `runTests` be in some other class that serves exclusively as a test runner.

Put another way, it would be better to put the _testing logic_ in its own class, so that it doesn't get mixed in with the logic for just running the analysis.

include/Utils.h
```diff
@@ -23,6 +23,8 @@ std::string variable(const Value *Val);
  */
 std::string address(const Value *Val);
 
+bool isNumber(const std::string& s);
```
kelloggm, line 26: Is this not in the standard library? I would think you could just call some version of `atoi`?

run_all.sh
```diff
@@ -8,7 +8,7 @@ echo "Making"
 eval "make" 
 
 
-for i in $(seq 1 14)
+for i in $(seq 1 15)
```
kelloggm, line 11: Would it be possible to automatically compute this number from the tests themselves, so that you don't need to update it each time?

Or, we could use an alternative design: give each test a unique name (rather than a number), and then run every test in the directory.

src/Analysis.cpp
```diff
@@ -374,380 +237,59 @@ void populateAliasedVars(Instruction* I, SetVector<Instruction *>& workSet, Alia
...
+
 
   if (!ALLOW_REDEFINE) {
         // check if code re-defines a safe function or memory function. if so, cast it as a unsafe function
     if (SafeFunctions[fnName]) {
       SafeFunctions[fnName] = false; 
       UnsafeFunctions[fnName] = true; 
       functionIsKnown = true; 
-      errs() << "CME-WARNING: Re-definition of safe function '" << fnName << "' identified on  and will be labelled as unsafe.\n";
+      errs() << "**ANALYSIS-WARNING**: Re-definition of safe function '" << fnName << "' identified on  and will be labelled as unsafe.\n";
```
kelloggm, line 279: I think there is something missing from this error message (after "on")

src/CalledMethods.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+            if (this->safeFunctions[fnName]) {
+                logout("safe func")
+                return; 
+            }
+
+
+            for (auto Pair : this->memoryFunctions) {
+                if (fnName == Pair.first) { // allocation function
+
```
kelloggm, line 133: Is this `if` block really necessary? It's empty.

kelloggm (changes requested):
This is definitely a dramatic improvement. I left a few comments (all I had time for today), but I will take a deeper look on Monday.

Another thing that we should do (not in this PR, but separately) is setup a code formatter to automatically format the code in this repo. I think my preferred choice would be [clang-format](https://clang.llvm.org/docs/ClangFormat.html), but I'm open to suggestions. Ideally, this would be configured so that it either runs automatically on commit, rejects non-formatted code in CI, or something similar (I don't have a strong preference).

include/Utils.h
Kevin-TD: I don't think there is; at most there is isdigit but nothing for an entire string

include/PassType.h
Kevin-TD: I opted for using maps since I'd imagine once we further build the code, there would be a lot of safe, unsafe, realloc, & memory functions to check for so a string->bool map would help with making lookup times faster

kelloggm: > I opted for using maps since I'd imagine once we further build the code, there would be a lot of safe, unsafe, realloc, & memory functions to check for so a string->bool map would help with making lookup times faster

The potential size of these maps shouldn't have an impact here: I'm asking about the data representation.

In particular, I don't understand the purpose of the `bool`. What does it represent? That is, when is it true and when is it false? It seems to me that a `std::set<std::string>` should be sufficient here: the presence of a function in the set means that it is a function of the named type; not being in the set means that it is not.

include/Utils.h
kelloggm: (You'd check its return value to determine if it succeeded, since it is a conversion function.)

kelloggm (changes requested):
This is a more-detailed review of `PassType` and `CalledMethodsPass`. I suspect that many of my comments also apply to `MustCallPass` given its similarity to `CalledMethodsPass`, so I'd like you to go through it after you've read my comments and fix anything that I mentioned in my review of `CalledMethodsPass` before I review it.

src/Analysis.cpp
Kevin-TD: Yeah this should just say "identifed and will be ..."

src/CalledMethods.cpp
```diff
@@ -0,0 +1,316 @@
...
+#define logOutMemory(x) printMemory(x);
+#else
+#define logout(x) 
+#define logDomain(x) 
+#define logOutMemory(x) 
+#endif 
+
+
+CalledMethods::CalledMethods(std::string passName) {
+    this->passName = passName;
```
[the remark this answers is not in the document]
Kevin-TD: Yeah this is just a place to hold the name of the pass in the test code (e.g., test1.txt) so the code knows whether a test case is for called methods or must call. I'm thinking of just putting it in a `Constants.h` file instead, like a `const std::string CALLED_METHODS_NAME = "CalledMethods"` and changing this constructor such that the constructor has no arg and it just does `this->passName = CALLED_METHODS_NAME`

src/CalledMethods.cpp
```diff
@@ -0,0 +1,316 @@
...
+    else if (auto Store = dyn_cast<StoreInst>(I)) {
+
+        Value* valueToStore = Store->getOperand(0);      
+        Value* receivingValue = Store->getOperand(1);   
+    
+        for (auto Inst : workSet) {
+          if (valueToStore == Inst) {
+
+            // check if it is allocation function 
+            if (auto Call = dyn_cast<CallInst>(Inst)) {
```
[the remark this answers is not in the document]
Kevin-TD: For the called methods pass itself, it does not need to do anything when an allocation function is called. Probably the reason why this comment is still here is that it's a remnant of code that previously checked for an allocation function that's now deleted. This'll be resolved on the next push where we have more generalizations for the transfer function

src/CalledMethods.cpp
```diff
@@ -0,0 +1,316 @@
...
+#define logDomain(x) 
+#define logOutMemory(x) 
+#endif 
+
+
+CalledMethods::CalledMethods(std::string passName) {
+    this->passName = passName; 
+}
+
+void CalledMethods::transfer(Instruction* I, SetVector<Instruction *> workSet, std::map<std::string, MethodsSetHolder>& inputMethodsSet) {
```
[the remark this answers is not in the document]
Kevin-TD: Agreed; it was a name that I picked up from the CIS lab code but I think as well it's better to change it to `instruction`

src/CalledMethods.cpp
```diff
@@ -0,0 +1,316 @@
...
+            
+
+            // * i THINK this is what distinguishes "real" variables defined in the program from constants and other irrelevant llvm stuff 
+            if (argName[0] != '%') return; 
+
+            std::list<std::string> allAliases;
+            allAliases.push_back(argName); 
+            while (argName != "") {
+                argName = this->aliasedVars[argName]; 
+                if (argName != "")
```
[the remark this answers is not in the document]
Kevin-TD: Agreed; will be doing that

src/CalledMethods.cpp
```diff
@@ -0,0 +1,316 @@
...
+                if (argName != "") 
+                    allAliases.push_back(argName); 
+            }
+
+            for (auto alias : allAliases) {
+                logout("alias = " << alias)
+            }
+
+
+            // * (i think this handles) cases where an entirely random and undefined (implicitly declared) function shows up
```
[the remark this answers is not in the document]
Kevin-TD: Pretty much; these are cases where the programmer might call a function that's never defined

include/PassType.h
```diff
@@ -0,0 +1,51 @@
...
+struct MethodsSetHolder {
+    std::set<std::string> methodsSet; 
+    bool setInitialized; 
+};
+
+typedef std::map<std::string, std::map<std::string, MethodsSetHolder>> MappedMethods;
+
+class PassType {
+private: 
+    AliasMap aliasedVars;
```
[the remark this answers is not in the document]
Kevin-TD: So `AliasMap` is a mapping between LLVM IR intermediate variables (e.g., %6, %7, %8) and variables visible in the code (e.g., %str). It's used to find all local must-aliases, though it is not expected to find all of them once we begin testing this code on more realistic, larger code 
A comment in the code explaining this will be added

include/PassType.h
```diff
@@ -0,0 +1,61 @@
...
+#define PASS_H
+
+#include "RunAnalysis.h"
+#include "Utils.h"
+#include "CFG.h"
+
+#include <set>
+#include <fstream>
+
+// mapping between LLVM intermediate variables (e.g., %6, %7, %8) and variables visible in the code (e.g., %str). Used to find all local must-aliases, though it is not expected to find all of them
```
kelloggm, line 11: "used to final **all** local must-aliases, though it is not expected to find **all** of them" (emphasis mine) is contradictory. I'd remove the first "all"

include/PassType.h
```diff
@@ -0,0 +1,51 @@
...
+typedef std::map<std::string, std::string> AliasMap;
+
+struct MethodsSetHolder {
+    std::set<std::string> methodsSet; 
+    bool setInitialized; 
+};
+
+typedef std::map<std::string, std::map<std::string, MethodsSetHolder>> MappedMethods;
+
+class PassType {
```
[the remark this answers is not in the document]
kelloggm: How about `DataflowPass`?

include/PassType.h
```diff
@@ -0,0 +1,61 @@
...
+
+// mapping between LLVM intermediate variables (e.g., %6, %7, %8) and variables visible in the code (e.g., %str). Used to find all local must-aliases, though it is not expected to find all of them
+typedef std::map<std::string, std::string> AliasMap;
+
+struct MaybeUninitMethodsSet {
+    std::set<std::string> methodsSet; 
+    bool setInitialized; 
+};
+
+// mapping for an entire program; first string is branch name, second string is var name
```
kelloggm, line 19: Why are branch names used as keys? This might be because I don't fully understand what LLVM is giving you, but it goes against my expectations that you'd use branch names at all to identify an estimate: I'd expect that there'd be an entry in this map for every program point, instead.

include/PassType.h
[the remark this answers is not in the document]
kelloggm: I'd also accept `GenericPass` or something similar to that if you'd prefer it, but I think `DataflowPass` is more specific.

run_all.py
```diff
@@ -0,0 +1,39 @@
+# run with `python3 ../run_all.py`  (must be in build dir)
```
kelloggm, line 1: Use a shebang instead

run_all.py
```diff
@@ -0,0 +1,39 @@
...
+
+import os
+import sys
+
+# checks if cwd is build
+if os.getcwd().split("/")[2] != "build":
```
kelloggm, line 7: This is definitely platform- and possibly machine-specific. Use `os.path.split(os.getcwd())[1]` instead.

src/PassType.cpp
```diff
@@ -0,0 +1,403 @@
...
+#include <fstream>
+
+void PassType::setFunctions(std::set<std::string> safeFunctions, std::set<std::string> unsafeFunctions, std::set<std::string> reallocFunctions, std::map<std::string, std::string> memoryFunctions) {
+    this->safeFunctions = safeFunctions;
+    this->unsafeFunctions = unsafeFunctions; 
+    this->reallocFunctions = reallocFunctions; 
+    this->memoryFunctions = memoryFunctions; 
+}
+
+void PassType::buildExpectedResult(std::string testName) {
```
kelloggm, line 12: I think this method logically belongs in `TestRunner`, not this class.

src/PassType.cpp
```diff
@@ -0,0 +1,403 @@
...
+
+void PassType::setCFG(CFG* cfg) {
+    this->cfg = cfg; 
+}
+
+void PassType::setAliasedVars(AliasMap aliasedVars) {
+    this->aliasedVars = aliasedVars; 
+}
+
+bool getAllowedRedefine(std::string testName) {
```
kelloggm, line 91: This method, too, logically seems to be part of `TestRunner` to me.

src/PassType.cpp
```diff
@@ -0,0 +1,403 @@
...
+        Value *argument = Call->getArgOperand(i);
+        std::string argName = dataflow::variable(argument);
+
+        argName = aliasedVars[argName];
+        while (argName.size() > 1 && dataflow::isNumber(argName.substr(1))) {
+          argName = aliasedVars[argName]; 
+        }
+
+
+        // * i THINK this is what distinguishes "real" variables defined in the program from constants and other irrelevant llvm stuff
```
kelloggm, line 192: Not needed in this PR, but we should open issues on this repo to resolve both this comment and the one on line 207. For both, the issue should suggest writing a test to check whether these are correct.

kelloggm (changes requested):
These comments are more minor than my last review: this code is much improved. Thanks! I think this should be ready to merge (hopefully) after one more round of changes.

run_all.py
Kevin-TD: What is a shebang?

include/PassType.h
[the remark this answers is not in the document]
Kevin-TD: I like `DataflowPass` so I'll go with that

include/PassType.h
Kevin-TD: I think here we can say branch names are program points; each branch name holds the must call/called methods estimate for each variable after all the code within that branch has run.

src/PassType.cpp
Kevin-TD: Upon testing and seeing what `argName` is at runtime I now realize that it's more apt to check if `argName` is an empty string (empty string meaning it's not in the `aliasedVars` map). `aliasedVars` filters out irrelevant LLVM IR data by virtue of only storing variables names and intermediate variables (e.g., %str, %1, %2). 
Now for the comment on line 207 I'm still unsure, so that'll probably get its own issue, but I don't think we'll need to go back on the comment for line 192 as I'm confident that we just need to check if `argName` is an empty string

run_all.py
kelloggm: https://en.wikipedia.org/wiki/Shebang_(Unix)

Basically, it's a directive to the OS that tells it what program to use to execute a script.

test/test15.c
```diff
@@ -6,33 +6,44 @@
...
 #define Annotate(x) __attribute__((annotate(x)))
 
-Annotate("DoesFree")
+#define MustCall(x) __attribute__((annotate("MustCall: {" #x "}")))
+#define CalledMethods(x) __attribute__((annotate("CalledMethods: {" #x "}")))
+#define MustCallUnkown __attribute__((annotate("MustCallUnknown: {}")))
+
+// unsure how much we'll be needing these 
+#define Owning __attribute__((annotate("Owning")))
+#define MustCallAlias __attribute__((annotate("MustCallAlias"))) // implementation will be tricky  since this needs a pair of annotations, one on the return type and one for a parameter
```
kelloggm, line 15: These should definitely be in their own header file that a client program (like a test) should include rather than in the tests directly. For now, this is probably fine, but we'll need to do that eventually so if it's easy to do you should do it now.

include/Utils.h
kelloggm: This conversation still hasn't been resolved, and should be before we merge.

kelloggm (approved):
Two very minor comments. If there's nothing controversial in their resolution, I do not need to approve again before this is merged.

include/Utils.h
Kevin-TD: So even if we use `strtol`, it wouldn't be a simple, short one-liner, hence why I still think defining a utility `isNumber` function is a good choice here. Since `isNumber` currently look likes: 
`char* endPtr;
  std::strtol(s.c_str(), &endPtr, 10);
  return endPtr != s.c_str() && *endPtr == '\0';`
  I'd rather have it remain a utility function with a descriptive name

kelloggm: Alright, thanks for looking into it. I agree that keeping a descriptive name is a worthwhile goal.

PR #8, opened 2023-07-09
Add clang format

Kevin-TD:
-file made that runs code formatting
-ci errors if code is not formatted 
-some unusued code removed
-annotations put in a new folder

### DIFF
--- a/format_code.py
+++ b/format_code.py
@@ -1,0 +1,7 @@
+#! python3 ../run_all.py
+# must be in /c-resource-leak-checker dir 
+
+import os
+import sys
+
+os.system("find . -name '*.cpp' -o -name '*.h' | xargs clang-format -style=LLVM -i")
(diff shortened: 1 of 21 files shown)


format_code.py
```diff
@@ -0,0 +1,10 @@
+#! python3 format_code.py
+# must be in ./c-resource-leak-checker
+
+import os
+
+# checks if cwd is build
+if os.path.split(os.getcwd())[1] != "c-resource-leak-checker":
+    print(f"WARNING: not in dir /c-resource-leak-checker; cwd is {os.getcwd()}")
```
kelloggm, line 8: I would expect this script to then exit, not just print a warning

kelloggm (approved):
Overall looks good. I don't need to review again once the one comment on the formatting script is resolved.

PR #9, opened 2023-07-26
Program variable

Kevin-TD:
replaces aliasmap with program variable type that handles names and aliases such that transfer functions are nicer


include/ProgramVariable.h
```diff
@@ -0,0 +1,45 @@
...
+private:
+  Value *value;
+
+  // name is llvm IR name without % or @
+  std::string cleanedName;
+
+  // name is llvm IR name maybe with % or @
+  std::string rawName;
+
+  std::list<ProgramVariable> aliases;
```
kelloggm, line 17: Use a comment here to specify which kind of aliases (may or must) these are, and whether the list is an over or underapproximation.

I think this is an underapproximation of must aliases (that's what the RLC for Java does). That is to say, every entry in this list is definitely an alias of the variable being represented, but there might be other aliases that are not in the list.

src/Analysis.cpp
```diff
@@ -19,6 +20,13 @@
 // note for run all tests is that if you add more tests, you'll have to modify
 // run_all.sh to include that test number
 
+// if a known function is being re-defined, issue a warning and do not check its
+// annotations. treat it as is we want to remove UnsafeFunctions and just have
```
kelloggm, line 24: This comment won't make sense once this PR is merged, I think, because you've removed all other mentions of `UnsafeFunctions` from the code. Can you rewrite it so that it makes sense without the reader needing to be familiar with the previous version of the code?

src/Analysis.cpp
```diff
@@ -252,60 +259,132 @@ void CalledMethodsAnalysis::doAnalysis(Function &F,
...
+                      dyn_cast<GlobalVariable>(note->getOperand(0))) {
+                if (ConstantDataSequential *data =
+                        dyn_cast<ConstantDataSequential>(
+                            annotateStr->getInitializer())) {
+                  if (expr->getOpcode() != Instruction::BitCast ||
+                      expr->getOperand(0) != &F) {
+                    continue;
+                  }
+                  logout("annotation = " << data->getAsString().str())
+                }
```
kelloggm, line 338: This code makes me very nervous, because it looks like you're fully deconstructing the annotations. That means that this code is going to be brittle to any changes in the program's structure or the annotation language; maybe that's okay for now.

Regardless, though, this code should be in its own helper function.

test/test15.c
```diff
@@ -1,86 +1,43 @@
 #include <stdio.h>
 #include <stdlib.h>
-#include <string.h>
+#include <unistd.h>
 #include "../Annotations/Annotations.h"
```
kelloggm, line 4: This test seems to have been completely changed. Can you explain the reasoning for that, rather than deleting this test and creating a new one (which would have a cleaner diff).

test/test16.c
```diff
@@ -0,0 +1,43 @@
...
+#include <stdlib.h>
+#include <unistd.h>
+#include "../Annotations/Annotations.h"
+
+struct my_struct {
+    char* x MustCall("free"); 
+    char* y MustCall("free"); 
+}; 
+
+void does_free(char* s CalledMethods("free")) {
```
kelloggm, line 11: This function looks surprising to me, because the annotation is not what I'd expect. In the Java RLC, `CalledMethods("free")` on a variable means that that `free()` has definitely already been called on that variable's value. But here, you seem to be using it to mean "free() **will** be called on this variable", which is not quite the same.

There is a similar annotation in the Java version: `EnsuresCalledMethods`, which is added to a method, means that the annotated method guarantees that one of its arguments (or a field, etc.) has a particular `CalledMethods` fact true about it when the method finishes.

kelloggm (changes requested):

include/ProgramVariable.h
Kevin-TD: It aliases the same way AliasMap did, and you're correct in that they're underapproximations of must aliases. I'll add that comment to the code

src/Analysis.cpp
Kevin-TD: Changing comment to "if a known function is being re-defined, issue a warning and remove it from safe/realloc/memory functions (wherever it's in). it's annotations should be checked (once annotations are fully implemented)"

src/Analysis.cpp
Kevin-TD: Can I move this to a helper function in a different pull request wherein I further build annotation handling?

test/test15.c
Kevin-TD: That was a mistake on my part. I'll add it back as test18

src/Analysis.cpp
kelloggm: Putting it into a helper function now is probably better, since we know we'll eventually want to do so.

test/test16.c
```diff
@@ -12,6 +12,12 @@ void does_free(char* s CalledMethods("free")) {
     free(s); 
 }
 
+/*
+TODO: change "CalledMethods" to EnsuresCalledMethods / Calls instead of "CalledMethods"
+- with EnsuresCalledMethods assume it's true & check it later
```
kelloggm, line 17: "later" isn't quite right: the distinction is between call sites and method definitions. For example, if you have:
```
Calls(x, "free")
void foo(void* x);
```
Then, at each call to `foo(y)`, we assume `y` has had `free()` called on it after `foo()` returns. At the definition of `foo()`, we have to prove that it actually does call `free` on `x`. Which happens first depends on the order in which the analysis visits the files (and the order is random, so we cannot rely on it either way).

kelloggm (approved):
I think we can merge this now, and then come back to fix the other things that we know are issues still.

PR #10, opened 2023-08-01
update to MustCall

Kevin-TD:
-MustCall now contains corresponding memory-freeing function instead of memory allocating function; test cases updated accordingly 
-some experimentation done with design of annotations; there's an implementation and design issue here


kelloggm:
> There are some issues with how LLVM IR handles annotations

These issues are serious. I agree that the irritation of writing the method name could have been tolerable, but that users will reject a tool that effectively requires them to specify information for each annotation twice: once with the annotation's location, and again with the annotation's arguments.

I am not a fan of your proposal for a .json file, for the simple reason that I think it will require users to do too much. I don't think we can reasonably expect an engineer to produce such a file for every piece of code that they're working with by hand - the overhead would just be too high.

Here are some alternative proposals:
* consider putting the annotations into comments and adopting a simpler annotation syntax. This would require us to be able to extract comment text and then associate it with particular variable declarations, but I suspect it's probably possible. The required engineering effort would likely be high for us, though. The original version of the Checker Framework used this strategy.
* add some kind of pre-processing step to simplify annotation *writing* but not *processing*. The key idea would be that users would write annotations without the "extra" positional information that we need (i.e., what function/struct/parameter/etc.), but that we'd use another analysis to recover that information from the program's structure. For example, if a user writes:
```
struct my_struct {
    MustCall(METHODS("free"))
    char* x; 
};
```
We would translate that to:
```
struct my_struct {
    MustCall(STRUCT("my_struct") AT_FIELD("x"), METHODS("free"))  // <--
    char* x; 
};
```
How? By observing that the `MustCall` annotation is placed on field `x` of struct `my_struct` during our pre-analysis pass.

We'd probably need to write that pre-analysis pass as some kind of separate analysis; I think we could do it just with a parse tree, but I need to think about it a bit.

Kevin-TD:
I like the idea of adding pre-processing to simplify annotation writing; additionally, we might be able to make the placement of the annotations more like how the RLC does it, as the annotations here were placed in order for LLVM's IR generator to pick it up (but evidently enough, it's not very reliable so it doesn't matter much here). 
But I am unsure how to go about implementation-wise (I have never worked with parse trees before). This is more of a heads-up on my lack of knowledge here

Kevin-TD:
Sounds good; if that's all, could you approve this merge request? This request is more to patch a bug with MustCall (stores calls to free instead of malloc now)

PR #11, opened 2023-08-03
introduces raw annotation string handling

Kevin-TD:


src/Annotation.cpp
```diff
@@ -0,0 +1,136 @@
...
+    annoType = AnnotationType::MustCallAnnotation;
+  }
+
+  if (annoType == AnnotationType::UndefinedAnnotation) {
+    return UndefinedAnnotation;
+  }
+
+  std::string methodsString;
+
+  for (int i = 7; i < chunks.size(); i++) {
```
kelloggm, line 68: I think the `7` here comes from the grammar given above: there are 7 required items, and then anything after that must be a method argument? If that's correct, you should add a comment to explain it (it wasn't obvious to me, and I had to figure that out by reading the code carefully).

kelloggm: If I'm wrong, then you definitely need a comment, so that I can understand it :)

src/Annotation.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+Annotation::Annotation() {
+  this->annotationType = AnnotationType::UndefinedAnnotation;
+}
+
+bool Annotation::annotationIsUndefined() {
+  return this->annotationType == AnnotationType::UndefinedAnnotation;
+}
+
+Annotation *generateAnnotation(const std::string &rawAnnotationString) {
```
kelloggm, line 17: Is it possible that this method will ever be called with an invalid `rawAnnotationString` (e.g., one that the user wrote incorrectly)? I think it might be possible, in which case we need to do error checking and then emit a useful error message rather than just segfaulting (which I think would be the current behavior, for example if `chunks` ends up having 4 or fewer elements due to an invalid input).

I think the best fix for this is to write code that validates that a raw string meets the method's assumptions into a `checkValid` method that returns `bool`. Then, call that method immediately when you read one of these strings from the user, so that you can safely make these assumptions here. Then, you can add a comment at the top of this method that refers to that one, which will re-assure someone reading this code that the e.g., unguarded array access to `chunks[4]` at line 78 is safe, because the input string is guaranteed to be in the proper format.

kelloggm (changes requested):

include/RunAnalysis.h
```diff
@@ -14,12 +14,16 @@
...
+
 #include <algorithm>
+#include <iostream>
 #include <iterator>
 #include <list>
 #include <map>
 #include <set>
+#include <sstream>
 #include <string>
+#include <vector>
```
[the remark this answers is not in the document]
Kevin-TD: Vectors are used in the `splitString` function in `Utils.cpp` and `splitString` is used in `Annotation.cpp`. I prefer to put all C standard library directives used in the `RunAnalysis.h` since practically all the .h files include `RunAnalysis.h` , but whether it's better style to put `#include <vector>` in Utils.h I am unsure of either

src/Analysis.cpp
```diff
@@ -288,10 +294,17 @@ void CalledMethodsAnalysis::doAnalysis(Function &F,
     }
   }
 
-  if (fnName != "main") {
+  if (!doAnnos) {
+    // annotation parsing is completed, next is
```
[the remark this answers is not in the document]
Kevin-TD: Yeah the last part of that sentence was supposed to be a segway for the TODOs; I'll change it to "next is the following TODO's "

src/Annotation.cpp
Kevin-TD: Yes you'd be correct; I'll add the following example as a comment to explain each part of `chunks` 
For the raw string "TOOL_CHECKER Calls target = FUNCTION(creates_obligation).PARAM(2).FIELD(x) methods = free1, free2"
  chunks = ['TOOL_CHECKER', 'Calls', 'target', '=', 'FUNCTION(creates_obligation).PARAM(2).FIELD(x)', 'methods', '=', 'free1,', 'free2']
  chunks[0] should be TOOL_CHECKER to help verify it's one of our own annotations and not an unrelated string
  chunks[1] is annotation type (Calls/MustCall)
  chunks[2] is "target" keyword, chunks[3] is "="
  chunks[4] is the target itself (FUNCTION/STRUCT potentially with sensible PARAM and FIELD specifiers)
  chunks[5] is "methods" keyword, chunks[6] is "="
  chunks[7..end] are all the methods

include/RunAnalysis.h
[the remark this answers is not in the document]
kelloggm: Can you add a comment explaining that these include directives are for all the .cpp files at the top of `RunAnalysis.h`? That would prevent my confusion in the future.

src/Annotation.cpp
kelloggm: Thanks, that sounds great - I think it would make this code a lot clearer.

include/ErrorAnnotation.h
```diff
@@ -0,0 +1,15 @@
+#ifndef ERROR_ANNOTATION_H
+#define ERROR_ANNOTATION_H
+
+// this is not an annotation for an error but effectively takes the place of an
+// error message used when we are checking if a raw annotation string is correct
+// and turns out it is not
```
kelloggm, line 6: When you write a comment like this one that includes something that a thing is and something that the thing is not, it's usually best to lead with the thing that it is. In this case, that means putting everything after the "but" in this comment first, and then adding as a note that this does not represent an error in the run-time sense.

kelloggm: The reason for this advice is that someone reading this file for the first time doesn't necessarily know the whole context. Instead, they are probably looking for an explanation for what this file **is** as opposed to what it **is not**. And we'd prefer to optimize comments for someone who is confused, since they're supposed to be explanatory :)

src/Annotation.cpp
```diff
@@ -0,0 +1,500 @@
...
+  }
+  // chunks[1] == "Calls" or chunks[1] == "MustCall"
+
+  if (chunks[2] != "target") {
+    logout("Annotation String Error 3: invalid annotation '"
+           << rawAnnotationString
+           << "' 'target' keyword missing or inappropriately placed; '"
+           << chunks[2] << "' found") return false;
+  }
+  // chunks[2] == "target"
```
kelloggm, line 112: What's the purpose of the "target =" keyword? If it's required that the target is next, could we simply remove it and check the next item?

One reason I might imagine you've included it is to make the annotations more readable. That's a good reason, but if so I'd argue that we should permit the "target" section and the "methods" section in either order.

src/Annotation.cpp
```diff
@@ -0,0 +1,500 @@
...
+           << rawAnnotationString
+           << "' malformed methods argument") return false;
+  }
+
+  // chunk[7..end] should just be method args
+
+  std::string methodsString;
+  for (int i = 7; i < chunks.size(); i++) {
+
+    // removing potential "\0"'s in the string
```
kelloggm, line 333: Why would there be `\0`s in the string?

src/Annotation.cpp
```diff
@@ -0,0 +1,500 @@
...
+  // TODO: issue an error if annotation is being re-defined
+
+  // we can assume string is correctly formatted
+
+  AnnotationType annoType;
+  std::vector<std::string> chunks =
+      dataflow::splitString(rawAnnotationString, ' ');
+
+  // raw annotation
+  // TOOL_CHECKER (Calls || MustCall) target = {FUNCTION ||
```
kelloggm, line 384: This comment should be higher up in the file: above the validation code, definitely.

src/Annotation.cpp
```diff
@@ -0,0 +1,500 @@
...
+            "Method Args Error 4 on '"
+            << rawMethodsString
+            << "': alnum character cannot be followed by space") return false;
+      }
+    }
+  }
+  return true;
+}
+
+bool rawStringIsCorrectlyFormatted(const std::string &rawAnnotationString) {
```
kelloggm, line 77: We should add a TODO to automate the tests you've written for this method in `test19.c`. I'm guessing you checked that test's output by hand to test this?

kelloggm (changes requested):

include/Utils.h
```diff
@@ -17,6 +17,21 @@ std::string variable(const Value *Val);
...
 
+std::vector<std::string> splitString(const std::string &input, char delimiter);
+
+void removeWhitespace(std::string &input);
+
+// inclusive bounds [i, j]
+std::string sliceString(const std::string &str, int i, int j);
+
+/*
+returns true if str begins with a letter or an underscore and is followed by by
```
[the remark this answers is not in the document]
Kevin-TD: What would the difference be in understanding what the function does if I put `iff` instead of `if`?

src/Annotation.cpp
Kevin-TD: It's more for readability on our end. But I don't think we should focus on being able to change the order of "target" and "methods" as we plan in the long run for the user only needing to be able to specify the methods.

src/Annotation.cpp
Kevin-TD: In the LLVM IR all strings end in "\00" automatically; I imagine this not being a problem once we write our own analysis on the C code but for now, we have to handle it

kelloggm (approved):

PR #12, opened 2023-08-08
Adds struct handling

Kevin-TD:


src/Analysis.cpp
```diff
@@ -346,6 +394,10 @@ void CalledMethodsAnalysis::doAnalysis(Function &F,
     doAnnos = true;
   }
 
+  if (fnName != "main") {
```
kelloggm, line 397: Am I correct in thinking that this early return means that the analysis won't run at all on methods not named "main"? If so, that's going to be a problem (probably sooner rather than later).

What's the reason for adding this?

test/test16.c
```diff
@@ -72,17 +55,42 @@ struct my_struct Calls("free", "x") does_something(struct my_struct S Calls("fre
...
+MustCall(STRUCT("my_struct") AT_FIELD("y"), METHODS("free"))
+MustCall(STRUCT("my_struct") AT_FIELD("x"), METHODS("free"))
+MustCall(STRUCT("my_struct") AT_FIELD("y"), METHODS("free"))
+Calls(FUNCTION("does_free") AT_PARAMETER("1"), METHODS("free"))
+Calls(FUNCTION("creates_obligation") AT_PARAMETER("1"), METHODS("free"))
+Calls(FUNCTION("creates_obligation") AT_PARAMETER("2") AT_FIELD("x"), METHODS("free"))
+MustCall(FUNCTION("creates_obligation") AT_RETURN, METHODS("free")) 
+Calls(FUNCTION("does_something") AT_RETURN AT_FIELD("x"), METHODS("free"))
+MustCall(STRUCT("my_struct") AT_FIELD("x"), METHODS("free"))
+MustCall(STRUCT("my_struct") AT_FIELD("y"), METHODS("free"))
```
kelloggm, line 68: These two are also duplicates with lines 58-59.

kelloggm (changes requested):

src/Analysis.cpp
Kevin-TD: That's correct; I have it here because I didn't need the entire analysis running on every single method as the test cases only focus on the main method. Moving on though, this line will be removed

src/ProgramVariable.cpp
```diff
@@ -74,5 +84,32 @@ std::set<std::string> ProgramVariable::getAllAliases(bool cleanNames) {
       allAliases.insert(pv.getRawName());
     }
   }
+  return allAliases;
+}
+
+std::list<ProgramVariable *> ProgramVariable::getNamedAliasesVars() {
```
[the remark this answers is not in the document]
Kevin-TD: The difference between `getNamedAliases, getAllAliases` and `getNamedAliasesVars, getAllAliasesVars` is just in the return type.  `getNamedAliases, getAllAliases` returns a set of strings representing its aliases, and `getNamedAliasesVars, getAllAliasesVars` returns a list of the original `ProgramVariable*`s representing the aliases; I added these methods because it could be useful to retrieve the original LLVM `Value*` stored inside a `ProgramVariable`, but I just realized I never actually used these new methods, so I might end up removing them. Also, could I add the comments to `ProgramVariable.h` instead?

test/test16.c
Kevin-TD: Yeah those are typos; I meant to refer to the other 2 structs

src/ProgramVariable.cpp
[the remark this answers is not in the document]
kelloggm: Adding the comment to `ProgramVariable.h` instead is fine.

For methods that aren't being used, I agree that it would be better to remove them.

include/ProgramVariable.h
```diff
@@ -42,11 +42,9 @@ class ProgramVariable {
 
   void addAlias(ProgramVariable pv);
 
+  // returns aliases as their string representation
```
kelloggm, line 45: I still don't feel like this comment explains the difference between these two seemingly-similar methods well enough for me to figure out exactly what I'm getting back. In both cases, you're returning `std::set<std::string>`, so isn't `getAllAliases` returning some string representation, too?

src/ProgramVariable.cpp
```diff
@@ -14,7 +14,7 @@ ProgramVariable::ProgramVariable(Value *value) {
 ProgramVariable::ProgramVariable(Value *value, int index) {
   this->value = value;
   this->rawName = dataflow::variable(value) + "." + std::to_string(index);
-  this->cleanedName = dataflow::variable(value) + "." + std::to_string(index);
+  this->cleanedName = this->rawName;
```
kelloggm, line 17: Shouldn't this be a true copy? I'm concerned that this is just copying the pointer, and that the `erase` call on line 20 below will also affect `rawName`.

include/ProgramVariable.h
Kevin-TD: So `getNamedAliases` returns the aliases that are named (i.e., are "actual" variables in the C code). Intermediate variables present in the IR like %7, %8 (that may be aliased to a named variable) are unnamed; variables present in the C code like %str are named. 
`getAllAliases` returns both named and unnamed aliases.

src/ProgramVariable.cpp
Kevin-TD: This makes a true copy. See https://cplusplus.com/reference/string/string/operator=/. If I were to copy the pointer it would look like `this->cleanedName = &this->rawName" and this->cleanName would be a string pointer

kelloggm: Great, sorry for my confusion. I'm slowly relearning C++ :)

include/ProgramVariable.h
kelloggm: That's a great explanation. Can you add it as a comment here? Then I think we are good to go on this PR.

kelloggm (approved):

PR #13, opened 2023-08-12
changes to testing suite + some annotation handling

Kevin-TD:
sorry for bundling so many disparate changes in one! i started working on annotation handling then realized that the testing suite really needed some updates

### DIFF
--- a/include/AnnotationHandler.h
+++ b/include/AnnotationHandler.h
@@ -1,0 +1,42 @@
+#ifndef ANNOTATION_HANDLER_H
+#define ANNOTATION_HANDLER_H
+
+#include "Annotation.h"
+#include "Debug.h"
+#include "ErrorAnnotation.h"
+#include "FunctionAnnotation.h"
+#include "ParameterAnnotation.h"
+#include "ReturnAnnotation.h"
+#include "StructAnnotation.h"
+#include "Utils.h"
+
+class AnnotationHandler {
+private:
+  std::list<Annotation *> annotations;
+
+public:
+  AnnotationHandler();
+  AnnotationHandler(const std::string &fileName);
+  void addAnnotation(const std::string &rawAnnotationString);
+  void addAnnotationsFromFile(const std::string &fileName);
+
+  // returns Annotation* instead of more specific type (in this case,
+  // FunctionAnnotation*) since we may not find the annotation and will want to
+  // return ErrorAnnotation* instead
+  Annotation *getFunctionAnnotation(const std::string &functionName);
+
+  Annotation *getParameterAnnotation(const std::string &functionName,
+                                     int nthParameter);
+  Annotation *getParameterAnnotation(const std::string &functionName,
+                                     int nthParameter,
+                                     const std::string &field);
+
+  Annotation *getReturnAnnotation(const std::string &functionName);
+  Annotation *getReturnAnnotation(const std::string &functionName,
+                                  const std::string &field);
+
+  Annotation *getStructAnnotation(const std::string &structName,
+                                  const std::string &field);
+};
+
+#endif
--- a/include/AnnotationHandler.h
+++ b/include/AnnotationHandler.h
@@ -10,6 +10,7 @@
 #include "StructAnnotation.h"
 #include "Utils.h"
 
+// manages annotations for a single file
 class AnnotationHandler {
 private:
   std::list<Annotation *> annotations;
@@ -23,6 +24,21 @@ public:
   // returns Annotation* instead of more specific type (in this case,
   // FunctionAnnotation*) since we may not find the annotation and will want to
   // return ErrorAnnotation* instead
+  /*
+  TODO: C++ has multiple inheritance, so we could also fix this issue by
+changing the type hierarchy. In particular, we could add types like
+FunctionErrorAnnotation that are both a FunctionAnnotation and an
+ErrorAnnotation; that is, make the type hierarchy look like this:
+
+         Annotation
+       /                   \
+ErrorAnno        FunctionAnno
+      \                     /
+    FunctionErrorAnno
+(and then we'd do the same for parameter annotations, etc.)
+
+
+  */
   Annotation *getFunctionAnnotation(const std::string &functionName);
 
   Annotation *getParameterAnnotation(const std::string &functionName,
--- a/include/ProgramVariable.h
+++ b/include/ProgramVariable.h
@@ -17,6 +17,12 @@ private:
   // under-approximation of must-aliases
   std::list<ProgramVariable> aliases;
 
+  // bool indicates whether the generated ProgramVariable is actually a program
+  // variable. It can be the case that the *Value is just the number 0, which
+  // isn't really the types of variables we are looking for. this is true if the
+  // LLVM name for the Value* starts with a % or @ and false if otherwise.
+  bool varIsIdentifier;
+
 public:
   ProgramVariable(Value *value);
 
@@ -41,6 +47,8 @@ public:
   bool equalsRawName(std::string otherRawName);
 
   void addAlias(ProgramVariable pv);
+
+  bool isIdentifier();
 
   // returns aliases as their string representation only for aliases that are
   // given names in the LLVM IR. intermediate variables like %7, %8 are unnamed;
--- a/include/ProgramVariable.h
+++ b/include/ProgramVariable.h
@@ -17,10 +17,21 @@ private:
   // under-approximation of must-aliases
   std::list<ProgramVariable> aliases;
 
-  // bool indicates whether the generated ProgramVariable is actually a program
-  // variable. It can be the case that the *Value is just the number 0, which
-  // isn't really the types of variables we are looking for. this is true if the
-  // LLVM name for the Value* starts with a % or @ and false if otherwise.
+  void fixNameAndIdentifier();
+
+  // bool indicates whether the generated ProgramVariable is actually an
+  // identifying variable in the program whose name in the IR begins with a % or
+  // @ or not.
+  /* this is relevant in the following case:
+  in the IR the instruction 'store i32 0, i32* %retval align` the alias
+  reasoning previously assumed that whatever we were storing into %retval would
+  be a variable/identifier and that we should store the 0 as a ProgramVariable
+  that's aliased to %retval, but it not correct to store the number zero as a
+  variable. additionally, if there's also the `instruction %0 = load i8**, i8***
+  %p, align 8, !dbg !23`, we should be aliasing %0 with %p and their cleaned
+  names would be 0 and p, but the program had would think the %0 was referring
+  to the ProgramVariable 0, messing up the aliasing reasoning
+  */
   bool varIsIdentifier;
 
 public:
--- a/include/ProgramVariable.h
+++ b/include/ProgramVariable.h
@@ -22,15 +22,11 @@ private:
   // bool indicates whether the generated ProgramVariable is actually an
   // identifying variable in the program whose name in the IR begins with a % or
   // @ or not.
-  /* this is relevant in the following case:
-  in the IR the instruction 'store i32 0, i32* %retval align` the alias
-  reasoning previously assumed that whatever we were storing into %retval would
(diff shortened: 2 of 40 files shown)


include/AnnotationHandler.h
```diff
@@ -0,0 +1,42 @@
...
+
+public:
+  AnnotationHandler();
+  AnnotationHandler(const std::string &fileName);
+  void addAnnotation(const std::string &rawAnnotationString);
+  void addAnnotationsFromFile(const std::string &fileName);
+
+  // returns Annotation* instead of more specific type (in this case,
+  // FunctionAnnotation*) since we may not find the annotation and will want to
+  // return ErrorAnnotation* instead
```
kelloggm, line 25: C++ has multiple inheritance, so we could also fix this issue by changing the type hierarchy. In particular, we could add types like `FunctionErrorAnnotation` that are both a `FunctionAnnotation` and an `ErrorAnnotation`; that is, make the type hierarchy look like this:
```
         Annotation
       /                   \
ErrorAnno        FunctionAnno
      \                     /
    FunctionErrorAnno
```

(and then we'd do the same for parameter annotations, etc.)

I'm not sure that this is worth it, though.

include/ProgramVariable.h
```diff
@@ -17,6 +17,12 @@ class ProgramVariable {
   // under-approximation of must-aliases
   std::list<ProgramVariable> aliases;
 
+  // bool indicates whether the generated ProgramVariable is actually a program
+  // variable. It can be the case that the *Value is just the number 0, which
```
kelloggm, line 21: by "just the number 0", do you mean that the `Value*` is nullable? Or that it's literally a reference to a zero, either in the IR or in the program itself?

src/Analysis.cpp
```diff
@@ -290,6 +303,33 @@ void doAliasReasoning(Instruction *instruction,
...
+      name]"
+      */
+      std::string structName = structType->getName();
+      structName = sliceString(structName, structName.find_last_of('.') + 1,
+                               structName.size() - 1);
+      int numFields = structType->getNumElements();
+      for (int i = 0; i < numFields; i++) {
+
+        Annotation *anno = annotationHandler.getStructAnnotation(
+            structName, std::to_string(i));
```
kelloggm, line 322: I'm worried that this might not work properly if there are two files being analyzed that each contain a struct that happens to have the same name. That case will come up in real code, so this needs to be keyed not just on the name of the struct, but also which file it is declared in (or something else that uniquely identifies a particular struct, which the name does not do).

src/Annotation.cpp
```diff
@@ -500,4 +510,16 @@ Annotation *generateAnnotation(const std::string &rawAnnotationString) {
...
\ No newline at end of file
+}
+
+std::string annotationTypeToString(AnnotationType anno) {
+  switch (anno) {
+  case AnnotationType::MustCallAnnotation:
+    return "MustCallAnnotation";
+    break;
+  case AnnotationType::CallsAnnotation:
+    return "MustCallAnnotation";
```
kelloggm, line 521: This doesn't seem right (copy-paste mistake?)

include/AnnotationHandler.h
```diff
@@ -0,0 +1,42 @@
...
+#include "Annotation.h"
+#include "Debug.h"
+#include "ErrorAnnotation.h"
+#include "FunctionAnnotation.h"
+#include "ParameterAnnotation.h"
+#include "ReturnAnnotation.h"
+#include "StructAnnotation.h"
+#include "Utils.h"
+
+class AnnotationHandler {
```
kelloggm, line 13: This class should have an explicit comment that indicates that it represents the annotations _in a particular file_ (based on reading the code, I think that's true). If it's not true, we're going to have problems with disambiguating names.

src/ProgramVariable.cpp
```diff
@@ -4,20 +4,26 @@
...
   this->value = value;
   this->rawName = dataflow::variable(value);
-  this->cleanedName = dataflow::variable(value);
+  this->cleanedName = this->rawName;
+  this->varIsIdentifier = true;
 
   if (this->cleanedName[0] == '%' || this->cleanedName[0] == '@') {
     this->cleanedName.erase(0, 1);
+  } else {
+    this->varIsIdentifier = false;
```
kelloggm, line 13: Since this code appears twice (here and at line 26 below), it makes sense to put it in a separate function (since it's more than a single line now).

src/TestRunner.cpp
```diff
@@ -49,68 +31,64 @@ bool TestRunner::runTests(MappedMethods expectedResult,
...
-          }
-          currentCM += c;
+      // we will be assuming our test case writing is correctly formatted.
+      // test arguments are written in one line and can be put in any order.
+      // they look like the following:
+      // pass=<str> function=<str> branch=<str> var=<str> methods=<set<str>>
+      // e.g., pass=MustCall function=main branch=entry var=s methods={free}.
+      // there must not be any spaces anywhere else, including in methods
+      // argument. to add more methods to method arg, it would look like:
+      // methods={x,y}.
```
kelloggm, line 57: If one of these comments is incorrectly formatted, we should at least try to fail politely - while writing test cases, we're all likely to make a mistake in formatting one of these occasionally, so it'd be best if that doesn't trigger a segfault.

src/Utils.cpp
```diff
@@ -1,23 +1,44 @@
 #include "Utils.h"
+#include "Debug.h"
 #include "RunAnalysis.h"
 
 const char *WHITESPACES = " \t\n\r";
 
 namespace dataflow {
 
+/*
+This code previously could not handle IR vars that arised from being a parameter
```
kelloggm, line 10: typo: "arised" -> "arose"

test/test20.c
```diff
@@ -0,0 +1,61 @@
...
+    free(S.x);
+    free(S.y); 
+}
+
+
+
+
+/*
+Right now I'd like the AT_PARAMETER and AT_FIELD to be 0-indexed 
+(I used 1-indexing before since I saw RLC use it with the field=#..., but just to be more consistent with LLVM I'll use 0-indexing) since if we had an annotation like AT_FIELD("x"),
```
kelloggm, line 44: Using 0-indexing is good. The RLC is 1-indexed to be consistent with the rest of the Checker Framework (the Java analysis tool that it is built on), which uses 1-indexing for parameters. (The reason that the CF uses 1-indexing for parameters is that most Java methods are instance methods of some object, so they implicitly have a "receiver" parameter---the object on which the method is called---which internally is stored as parameter 0. So, the RLC is sort of 0-indexed, too, but that's not visible externally.)

kelloggm (changes requested):
You can split up PRs that are unrelated by using `git` to only commit the parts that you want to be reviewed together. That is, rather than running `git commit -a`, which adds all of your changes to the commit, you can explicitly include files with `git add`; in that way, you could have separated out the test changes in this PR to avoid the potpourri feeling that you refer to in your PR description.

include/AnnotationHandler.h
kelloggm: In that case I'd suggest adding a todo comment (maybe just copy what I wrote above?)

include/ProgramVariable.h
Kevin-TD: It's literally the number zero; this code was written when I saw in the IR the instruction `store i32 0, i32* %retval, align 4`. My alias reasoning initially assumed that whatever we were storing into `%retval` would be a variable/identifier and that we should store the 0 as a ProgramVariable that's aliased to `%retval`, but it not correct to store the number zero as a variable.
I found this assumption quite troubling when in that same IR, the instruction `%0 = load i8**, i8*** %p, align 8, !dbg !23` was present. My code finds aliases between variables by looking at cleaned (no % or @ as the first character) variable names. For this instruction, we would be aliasing %0 with %p and their cleaned names would be 0 and p. But the program had mistakenly thought the %0  was referring to the ProgramVariable 0, messing up the aliasing reasoning.

src/Annotation.cpp
Kevin-TD: Yep; this'll be fixed

include/AnnotationHandler.h
Kevin-TD: When you had asked me about it in the meeting I hadn't thought about whether this was going to be for a file or for a program, but after thinking about it more, it seems best that it'll be for each file. I'll start writing tests that analyze several files in a directory

src/Utils.cpp
```diff
@@ -1,23 +1,44 @@
...
+name] = [inst]. but for parameters, it is just [type] [name] (e.g., "i8* %s"
+instead of something like "%s = alloca i8*, align 8").
+*/
 std::string variable(const Value *Val) {
   std::string Code;
   raw_string_ostream SS(Code);
   Val->print(SS);
+
+  // checks if Code does not contain '=', in which case, treat it as a parameter
+  // to a function
```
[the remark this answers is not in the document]
Kevin-TD: Typically, variables are always initialized by some alloca instruction. So `int i;` would look like `%i = alloca i32`. Even if it were `int i = 1;` instead, we'd first see `%i = alloca i32` and then the instruction `store i32 1, i32 %i`. This is at least the case for when we define new variables inside the body of a function. The case where an '=' did not show up was in this IR snippet: 

```
; Function Attrs: noinline nounwind uwtable
define dso_local void @does_free(i8* %s) #0 !dbg !10 {
entry:
  %s.addr = alloca i8*, align 8
  store i8* %s, i8** %s.addr, align 8
```
Here, at the store instruction, when we try getting the string name of %s, the Code looks like `i8* %s` instead of  something like `%s = alloca i8*`. I am assuming this is the case since %s is a parameter and not a newly defined variable.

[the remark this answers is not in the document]
kelloggm: Thanks for the explanation. I think a comment with this information in the code would be helpful here.

include/AnnotationHandler.h
kelloggm: My only concern about having an `AnnotationHandler` for each file is about `#include` directives: if a header file with annotations is included in another file that's being analyzed, will the annotations show up in the `AnnotationHandler` for the file under analysis, or will it need to be linked to the header file's `AnnotationHandler` somehow?

include/ProgramVariable.h
kelloggm: Thanks, that makes sense. I think the comment you wrote in the code here is a bit confusing, so moving at least some of this explanation into the comment in the code would be helpful.

include/AnnotationHandler.h
Kevin-TD: I have found out that the annotations do not show up in the IR of the file that's being analyzed. This is actually quite a problem since if we call some annotated function that's from some other .h file, there are 0 hints in the IR that even indicate where the function comes from. The #include directives aren't even listed in the IR so we'd have to look at the .c file itself to find them. I have been researching to see if there's some way to include everything in the #include directives into the IR for some files to no success. I'm quite unsure how this should be handled, so what do you suggest?

kelloggm: Yes, this is a very serious problem. I'm not sure what the "right" way to do this in C is.

Here is one possible solution:
* first, the analysis makes a quick (AST only) pass over all input files, only looking for annotations. It creates an internal model (using something like `AnnotationHandler`) of every annotated function in the program.
* then, we run our existing analysis.

We could also separate out the first part as a separate program that produced "annotation files" or something similar that our analysis knows how to read. I'm not sure that has any real benefits, though.

Kevin-TD: I think with this AST pass we'll also have to see what the files include as well since there's still the issue of file disambiguation. I like the idea of generating an annotation file for our analysis to read; I'm thinking it could tell us, for some annotation, which files that annotation applies to so that we know we're applying the correct one when we are analyzing a specific file.

kelloggm: "AST" refers to an [Abstract Syntax Tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree), which is the output of a parser. Some part of LLVM should be able to produce one from C code, though I'm not sure which part.

PR #14, opened 2023-08-23
CVE-2022-3544 test case

kchen2024:


kelloggm:
@Kevin-TD can you please review this test case that @kchen2024 has built? Thanks!

test-cases/CVE-2022-3544.h
```diff
@@ -0,0 +1,319 @@
+// this is the header file for CVE-2022-3544
+//#pragma once
+#include <malloc.h>
+#include <stdbool.h>
+//#include "damon_operations.h"
```
Kevin-TD, line 5: How come there are comments here for //#pragma once and //#include "damon_operations.h"? If they are just here for preservation, you can definitely remove them now as it'll be saved on GitHub. If not, add a comment why these include directives are commented out

PR #16, opened 2023-09-30
Annotation generation developed

Kevin-TD:
-code can now generate annotations properly with the user not needing to specify as much. testing framework for annotation generator required, though 
-deleted test18; not sure what was being tested. i don't know what it's useful for 
-deleted test19; poor way of annotation testing, as much less specifications are now required


src/Annotations/Annotation.cpp
```diff
@@ -95,32 +101,40 @@ handling for it is here just in case
...
 'FUNCTION(creates_obligation).PARAM(2).FIELD(x)', 'methods', '=', 'free1,',
-'free2'] chunks[0] should be TOOL_CHECKER to help verify it's one of our own
-annotations and not an unrelated string chunks[1] is annotation type
-(Calls/MustCall) chunks[2] is "target" keyword, chunks[3] is "=" chunks[4] is
-the target itself (FUNCTION/STRUCT potentially with sensible PARAM and FIELD
-specifiers) chunks[5] is "methods" keyword, chunks[6] is "=" chunks[7..end]
-are all the methods
+'free2']
+
+chunks[0] should be TOOL_CHECKER to help verify it's one of our own
```
kelloggm, line 128: add "the literal string" and some quotes here to make it clear that `TOOL_CHECKER` is not intended to be a variable: `chunks[0] should be the literal string "TOOL_CHECKER" to help ...`

(Aside: should we choose a less-generic name than TOOL_CHECKER?)

Annotations/annotation_generator.py
```diff
@@ -0,0 +1,392 @@
...
+
+    if ".FIELD" in anno_unfilled_target:
+            anno_unfilled_target_field_name = anno_unfilled_target[
+                anno_unfilled_target.find("(") + 1 : anno_unfilled_target.find(")")
+            ]
+            
+            if type(spec) is Function:
+                if param_index is None:
+                    logout(f"ret type is '{spec.return_type}'")
+                    # Assumption: spec.return_type looks like 'struct <struct_name> '
```
kelloggm, line 259: Why can't `spec.return_type` be a primitive (e.g., `int`)?

Annotations/annotation_generator.py
```diff
@@ -0,0 +1,392 @@
+#! python3 annotation_generator.py <ast txt file> <output for annotations file> 
+# generate AST text with: clang -Xclang -ast-dump -fsyntax-only -fno-color-diagnostics <c file> > <ast txt file>
+
```
kelloggm, line 3: This code is complex enough that you should add some tests for it. A few shell scripts that execute this script on the tests (using the command lines below) and compare its results to the expected output should be sufficient; just make sure that they run in CI.

src/Analysis.cpp
```diff
@@ -137,6 +136,61 @@ void buildCFG(CFG &topCFG, std::vector<std::string> branchOrder,
   }
 }
 
+std::vector<std::string> getAnnotationStrings(std::string optLoadFileName) {
+  std::string testName = getTestName(optLoadFileName);
+  std::string parentName = getParentName(optLoadFileName);
```
kelloggm, line 141: Please document the format of `optLoadFileName`. I'm not sure what the purpose of this code is, since I would expect our tool to take a list of `.c` files as input.

(Related question: have you tested this on `.h` files? We should add at least one test with an annotation in a header file.)

kelloggm (changes requested):
Overall this looks like a great improvement! With a few additions I think it will be great.

src/Annotations/Annotation.cpp
Kevin-TD: > should we choose a less-generic name than TOOL_CHECKER? 

How is `C_RESOURCE_LEAK_CHECKER_ANNOTATION`? Not sure what a good, less-generic name should be. Nonetheless, I think I'll be removing the 'TOOL_CHECKER' part entirely. I added it in the first place since how I first retrieved annotations was by looking at the global annotations generated in an LLVM IR file, and sometimes there would be unrelated strings. But since we now get the annotations by the AST and store them in a tmpfile, the 'TOOL_CHECKER' part is really not needed

kelloggm: Removing it entirely sounds good to me.

Annotations/annotation_generator.py
Kevin-TD: Adding the following comment: 
```
               # Assumption: spec.return_type looks like 'struct <struct_name> '
                    # we can make this assumption because: 
                    # 1. param_index is none, which means we are making an 
                    # annotation for the function itself, and (for now) we assume
                    # that means it's an annotation on its return.
                    # note: this will need to update
                    # if/when annotations like @Owning and @MustCallAlias are implemented
                    # 2. the ".FIELD" in anno_unfilled_target is specified by the user via 
                    # something like 
                    '''
                    char *MustCall("free") 
                        creates_obligation(char *s Calls("free"), struct my_struct X Calls("free", "x"));
                                                                                                   ^^^
                    '''
                    # the calling of that second parameter on Calls/MustCall tells us that for the struct 
                    # we are specifying, we are targeting some field. we verify whether that field 
                    # actually exists, since this pass does
                    # keep note of every struct and its fields. if we wrote something like:
                    # void Calls("free", "x") example(int S Calls("free", "x")) { ... }
                    #                                 ^^^                 ^^^ implying "int" has field "x"
                    # we'll throw "ValueError: Did not find field 'x' for struct 'int', anno TOOL_CHECKER Calls target = _.FIELD(x) methods = free"
```

src/Analysis.cpp
Kevin-TD: Turns out getTestName and getParentName are not necessary at all. I also kind of forgot why I had them in the first place. I think it was because I assumed `optLoadFileName` specified a `.ll` file, but after testing it out, it just specifies a `.c` file, so no extra parsing is needed. 

> have you tested this on .h files? 

If you mean where I directly feed in a `.h` file, no.  The command we run to test a .c file is just: 

```
command="make ; clang -emit-llvm -g -S -fno-discard-value-names -Xclang -disable-O0-optnone -c ../test/test$1.c -o ../test/test$1.ll ; opt -load CodeAnalyzer.so -CodeAnalyzer ../test/test$1.ll"
```
where `$1` is the number of the test (e.g., test1, test2, ...). Clang cannot generate IR for header files. I have tried feeding it a header file and it puked unparsable weird text encodings on my screen. I do have tests such as `test/test23/test23.c` and `test/test23/test24.c` that include header files with annotations from `test/test23/test23.h`, and those annotations do appear in the AST for the `.c` files.

kelloggm: >  I do have tests such as test/test23/test23.c and test/test23/test24.c that include header files with annotations from test/test23/test23.h, and those annotations do appear in the AST for the .c files.

Thanks, that's definitely sufficient.

Annotations/annotation_generator.py
kelloggm: This looks good, with one note:

> # note: this will need to update
> # if/when annotations like @Owning and @MustCallAlias are implemented

Please describe what update will be needed.

kelloggm (approved):
LGTM

PR #17, opened 2023-10-27
Annotation Reading Implementation

Kevin-TD:
-reads annotations and applies MustCalls/CalledMethods to the appropriate variables
-field member on annotations changed to number instead of string
-adds handling to a compiler optimization (sret attributes on structs) 
-adds handling to struct pointers


src/Analysis.cpp
```diff
@@ -279,13 +281,48 @@ void doAliasReasoning(Instruction *instruction,
...
+          dataflow::unwrapValuePointerToStruct(receivingValue);
+
+      if (valueStruct && receivingStruct) {
+        logout("two structs to alias " << *store);
+        int numFields = valueStruct->getNumElements();
+
+        // ASSUMPTION: both structs are referring to the same struct (or the
+        // structs have the same fields). note that i do not have strong reason
+        // to believe that they could be different, as the instruction is
+        // storing one struct into another.
```
kelloggm, line 305: In general I don't think this is a safe assumption to make. C compilers generally will permit the programmer to assign a value of any struct type to any other struct type.

I think we need to write code here to conservatively handle the case where the structs have completely different layouts. In that case, we should generally try to just assume that anything could happen: that is, "go to top" and drop as much information as possible.

kelloggm (commented):
This was a very surface-level review. I'll try to give a more in-depth review on the new code tomorrow some time.

src/Annotations/ReturnAnnotation.cpp
```diff
@@ -2,42 +2,32 @@
...
   }
 
   return "@" + annoTypeString +
          " ReturnAnnotation Function Name = " + this->targetName + " " +
          fieldString + " methods = " + annoMethodsString;
 }
 
-bool ReturnAnnotation::fieldNameEquals(const std::string &field) {
-  return field.compare(this->field) == 0;
+bool ReturnAnnotation::fieldNameEquals(int field) {
```
kelloggm, line 29: This no longer checks name equality, so the name of the method should change, too.

src/DataflowPass.cpp
```diff
@@ -29,68 +29,107 @@ void DataflowPass::setCFG(CFG *cfg) { this->cfg = cfg; }
 void DataflowPass::transfer(Instruction *instruction,
                             SetVector<Instruction *> workSet,
                             ProgramPoint &inputProgramPoint) {
-
+  // TODO: remove workset from arg list
```
kelloggm, line 32: Can you expand this comment just a little bit to expand on why this is desirable. Just a "because ..." and then a reason is enough.

src/DataflowPass.cpp
```diff
@@ -108,6 +147,7 @@ void DataflowPass::transfer(Instruction *instruction,
     for (unsigned i = 0; i < Call->getNumArgOperands(); ++i) {
       ProgramVariable argumentVar = ProgramVariable(Call->getArgOperand(i));
       std::string arg = argumentVar.getCleanedName();
+      ProgramVariable *pv = inputProgramPoint.getPVRef(arg, false);
```
kelloggm, line 150: This isn't used until over 20 lines later, at line 172. Is there a good reason to compute it so early? I don't see one immediately.

src/DataflowPass.cpp
```diff
@@ -168,79 +205,132 @@ void DataflowPass::transfer(Instruction *instruction,
...
+          MethodsSet *methods = pv->getMethodsSetRef();
+          logout("GETMAIN-5 ON '" << arg << "'");
+          this->onDeallocationFunctionCall(methods, fnName);
           loopBroken = true;
           break;
         }
       }
 
-      if (loopBroken) {
+      if (loopBroken || fnName == "llvm.dbg.declare") {
```
kelloggm, line 222: This extra condition definitely needs a comment to justify it. I think this might be debugging code, and if so it should also be guarded by a debug flag.

src/DataflowPass.cpp
```diff
@@ -168,79 +205,132 @@ void DataflowPass::transfer(Instruction *instruction,
...
       }
 
-      if (fnName == "llvm.dbg.declare") {
-        continue;
+      // finds annotations
+      logout("finding annotations for " << fnName);
+
+      /*
+        LLVM uses sret attributes as an optimization if the return type is
+        large. a struct with more than 2 fields can trigger this optimization.
```
kelloggm, line 231: I think "more than 2" here can also mean "any array or struct field"

kelloggm: which technically might just be one (the array/struct), but it has more than a single element

src/DataflowPass.cpp
```diff
@@ -168,79 +205,132 @@ void DataflowPass::transfer(Instruction *instruction,
...
+        (this is from C code
+        my_struct A1 = does_something_a(A);)
+
+        *we assume LHS is always a struct and first argument always has sret
+        attribute
+
+        we will:
+        -destructure A1 into its fields (A1.0, A1.1, ...)
+        -determine A1's MustCall/CalledMethods from annotations on return on
+        does_something_a -destructure A into its fields (A.0, A.1, ...)
```
kelloggm, line 256: I think there's a missing newline in this comment. Otherwise, it doesn't really make sense.

src/DataflowPass.cpp
```diff
@@ -168,79 +205,132 @@ void DataflowPass::transfer(Instruction *instruction,
...
+        %struct.my_struct* byval align 8 %A), !dbg !58
+
+        here's how we will handle these types of instructions:
+
+        say we have the instruction
+        call void @does_something_a(%struct.my_struct* sret %A1,
+        %struct.my_struct* byval align 8 %A), !dbg !58
+
+        (this is from C code
+        my_struct A1 = does_something_a(A);)
```
kelloggm, line 248: Please use better names than "A1" vs "A": this is hard enough to keep track of as it is :)

Maybe "A" and "B"?

src/DataflowPass.cpp
```diff
@@ -168,79 +205,132 @@ void DataflowPass::transfer(Instruction *instruction,
...
+    if (!structType && allocate->getAllocatedType()->isPointerTy()) {
+      structType = dataflow::unwrapValuePointerToStruct(allocate);
+    }
+
+    if (!structType) {
+      return;
+    }
+
+    std::string structName = structType->getName();
+    structName = dataflow::sliceString(
```
kelloggm, line 314: Minor, but I think `sliceString` belongs in Utils rather than dataflow.

kelloggm: Probably can/should be addressed in a separate, smaller PR

src/ProgramRepresentation/ProgramPoint.cpp
```diff
@@ -14,24 +14,77 @@ ProgramPoint::ProgramPoint(std::string pointName, ProgramPoint *programPoint) {
 
 void ProgramPoint::addAlias(ProgramVariable receiving,
                             ProgramVariable receiver) {
+  std::string addedAliasRawName;
```
kelloggm, line 17: Is the value stored in this variable every used? I don't think it is.

kelloggm (changes requested):
This PR was especially difficult to review because substantive changes were mixed with changes to formatting. In the future, please follow the best practice and put formatting changes (esp. from an autoformatter) in their own PR.

Testers/Annotations/test7/test7.txt
```diff
@@ -1,8 +1,8 @@
-MustCall target = FUNCTION(malloc0).RETURN methods = free
-MustCall target = FUNCTION(malloc1).RETURN methods = free
-MustCall target = FUNCTION(malloc2).RETURN methods = free
-MustCall target = FUNCTION(malloc3).RETURN methods = free
-Calls target = FUNCTION(free0).PARAM(0) methods = free
-Calls target = FUNCTION(free1).PARAM(0) methods = free
-Calls target = FUNCTION(free2).PARAM(0) methods = free
-Calls target = FUNCTION(free3).PARAM(0) methods = free
```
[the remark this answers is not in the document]
Kevin-TD: When I update the annotations in some test `.c` file, I have to update the `.txt` file. These annotations are generated by the code, but I first check for accuracy and then use them as the expected result. It's similar to how the passes in `Testers/Passes` are manually written text files. 

When the annotations are tested, we re-generate the annotations locally and do a simple `diff` between that and the `.txt` file.

include/Annotations/ParameterAnnotation.h
```diff
@@ -5,22 +5,25 @@
 
 class ParameterAnnotation : public Annotation {
 private:
-  std::string field;
+  // equals -1 if not defined
```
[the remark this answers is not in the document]
Kevin-TD: Exactly right! I'll add that comment

include/Annotations/ReturnAnnotation.h
```diff
@@ -5,19 +5,22 @@
 
 class ReturnAnnotation : public Annotation {
 private:
-  std::string field;
+  // equals -1 if not defined
+  int field;
+
+  // returns true iff field does not equal -1
   bool hasField;
```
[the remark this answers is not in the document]
Kevin-TD: I had it there before to store the result of doing (psuedocode) `fieldName is not empty`, which I thought made more sense to do since the field was a string, but I can see now we don't need it. I'll remove `hasField`

src/Analysis.cpp
```diff
@@ -343,36 +382,83 @@ void doAliasReasoning(Instruction *instruction,
...
+
+    StructType *structType =
+        llvm::dyn_cast<llvm::StructType>(allocate->getAllocatedType());
+
+    if (!structType && allocate->getAllocatedType()->isPointerTy()) {
+      structType = dataflow::unwrapValuePointerToStruct(allocate);
+    }
+
+    if (!structType) {
+      return;
```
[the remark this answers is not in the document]
Kevin-TD: Yes, it is possible for a primitive to be there. For example: 
```
int a = 1;
int* b = &a; 
```
For the line `int* b = &a`, it corresponds to the allocate instruction `%b = alloca i32*, align 8`. In this section of `Analysis.cpp`, we first check the allocation type to see if it's a struct type. It's not, but it's a pointer type, so we unwrap it to see if there is a struct type. In our unwraping, see we can't cast it to a struct type, so `structType = NULL`. Thus, we must early return. 

I realize that putting in this extra check unnecessarily complicates, though, so I'll simplify it to: 

```
 StructType *structType = dataflow::unwrapValuePointerToStruct(allocate);

    if (!structType) {
      return; 
    }
```

src/Analysis.cpp
```diff
@@ -343,36 +382,83 @@ void doAliasReasoning(Instruction *instruction,
...
+    }
+
+    programPoint->addVariable(ProgramVariable(allocate));
+
+    /*
+    I cannot find documentation that states how struct names appear in the IR,
+    but it seems that they typically look like "struct.[C struct name]"
+    TODO: prove/disprove that struct names are formatted as "struct.[C struct
+    name]"
```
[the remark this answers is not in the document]
Kevin-TD: Turns out, there is no required format, but it mostly looks like "struct.[C struct name]." I'm not sure how else it could be formatted (I didn't see any other types of formatting in the docs), so to compromise, I'll use the debug information. There are DICompositeType tags in the debug information that specify a type and a name. The name tag in DICompositeType tags are more reliable and likely to align with the C code. So for each struct name `structName` we find, we'll search the debug info to see if there is a DICompositeType that specifies a struct with the name `structName`. If we don't find any, we'll error and adjust the code to handle it. 

I found one case where the struct formatting is not "struct.[C struct name]", and that was by manually editing the IR. Users will definitely not be editing the IR, so I won't worry about this case.

src/Analysis.cpp
```diff
@@ -343,36 +382,83 @@ void doAliasReasoning(Instruction *instruction,
...
+    */
+    std::string structName = structType->getName();
+    structName = sliceString(structName, structName.find_last_of('.') + 1,
+                             structName.size() - 1);
```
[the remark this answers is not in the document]
Kevin-TD: `structName` will now have a use beyond debugging; see the conversation about "struct.[C struct name"

src/DataflowPass.cpp
Kevin-TD: `workSet` is not used. I'll just implement the change now

src/DataflowPass.cpp
Kevin-TD: The same `pv` ref is used in many other spots in the scope it's defined in, so I'd rather initialize `pv` then refer to it when we need it, rather than repeatedly initialize it in all the inner scopes. The difference would pretty much be doing this: (pseudocode)  

```
pv = input(M) 

if (a) { 
  use1(pv)
  continue
}

if (b) { 
  use2(pv)
  continue
}

```

instead of this: 
```
if (a) { 
  pv = input(M) 
  use(pv)
  continue
}

if (b) { 
  pv = input(M)
  use(pv)
  continue
}

```
and I prefer the former design. But, let me know if I actually misinterpreted your comment, or if the latter design is preferable.

src/DataflowPass.cpp
Kevin-TD: It's not debugging code; it's just that llvm.dbg.declare functions get lumped in as function calls along with function calls we make in the C code. llvm.dbg.declare are calls generated to declare debug info within the IR. 

At this point, we're looking for functions with annotations, and llvm.dbg.declare will not be a function with annotations, so we'll not check it. I'll add a comment noting this.

src/DataflowPass.cpp
Kevin-TD: I'm thinking of just having a `Utils` and `dataflow` namespace instead of only `dataflow`. It seems minor enough to change it in this PR, though.

src/ProgramRepresentation/ProgramPoint.cpp
Kevin-TD: It is; see lines 43–87 of this function

kelloggm: Ah, sorry. I think when I was reviewing and left this comment I had mistakenly read this function as ending at line 41, rather than continuing past that long comment.

src/DataflowPass.cpp
kelloggm: > It seems minor enough to change it in this PR, though

This is a dangerous impulse that leads to PRs that are hard to review, because they contain many unrelated changes. In general, anything that can be a small, simple PR should be, so that it can be reviewed and merged quickly and easily.

src/Analysis.cpp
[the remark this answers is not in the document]
kelloggm: This sounds like a good solution. Please make sure to update the comment in the code to explain this, too.

src/DataflowPass.cpp
kelloggm: I see what you mean. Fundamentally, the issue here is really just that this method is too big: it doesn't all fit onto a screen, and so to read it I have to constantly scroll up and down. This means that it's hard to see where variables come from when they are used.

For methods like this, I usually try to write helper methods for the parts inside the various `if` statements. So, the psuedocode would become:
```
pv = input(M) 

if (a) { 
  aLogic(pv, ...);
}

if (b) { 
  bLogic(pv, ...);
}
```

This lets a reader see the high-level structure easily, and then the reader can dive into each part on its own.

Testers/Annotations/test7/test7.txt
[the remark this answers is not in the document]
kelloggm: (Thanks for explaining, and sorry that I'm easy to confuse with these kind of minor details. There are a lot of them to keep track of, between all of the projects that I do reviews for!)

src/Analysis.cpp
Kevin-TD: This assumption is now dropped, and if they differ, we warn that aliasing information is lost.

src/Annotations/AnnotationHandler.cpp
```diff
@@ -72,12 +72,46 @@ AnnotationHandler::getParameterAnnotation(const std::string &functionName,
...
+    }
+  }
+
+  return annotationsWithFields;
+}
+
+std::vector<Annotation *>
+AnnotationHandler::getAllParameterAnnotationsWithoutFields(
+    const std::string &functionName) {
+  std::vector<Annotation *> annotationsWithFields;
```
kelloggm, line 95: The name of this variable is misleading (probably copy-paste error).

test/diff_type_struct_ptr/test__.c
```diff
@@ -0,0 +1,28 @@
+#include <stdio.h>
```
kelloggm, line 1: Very weird filename! In general, it would be nice to change the names of these tests so that they briefly indicate what they are testing (i.e., they aren't just "test1.c", "test2.c", etc., but instead the names have a tiny bit of semantics). As it is, it'll be hard to find the tests for particular features later (although the test directories definitely help!).

kelloggm (approved):
Overall I think this is mergable without another review from me. I still think it could be improved, but it's too large for me to suggest more fine-grained changes. I might do so as PRs (if I can find the time, haha).

test/diff_type_struct_ptr/test__.c
Kevin-TD: Ah, you were not supposed to see these files! This "test" was me playing with struct pointers of different types, which made me conclude that I won't be aliasing them. I remembered to not push these changes at first, but I forgot to do that this time. I'll put tests that are just for me in a different folder and is .gitignore'd

PR #18, opened 2023-11-15
Code cleanup & re-organization 

Kevin-TD:


kelloggm:
@Kevin-TD is this meant to be reviewed before #16 is reviewed again? I expected it to contain only non-semantic changes (i.e., to formatting etc.) based on the title, but it appears to have large overlap with #16. I'm going to re-review that one now, and I'll return to this one tomorrow after you've had a chance to clarify.

kelloggm:
Oops, that's a typo in my previous comment. I meant #17, not #16.

Kevin-TD:
This was meant to be reviewed after #17.

include/RunAnalysis.h
```diff
@@ -39,7 +39,7 @@
 
 using namespace llvm;
 
-namespace dataflow {
+namespace Dataflow {
```
kelloggm, line 42: C++'s convention is all lowercase for namespaces, so I don't think we should make this change. See e.g. https://google.github.io/styleguide/cppguide.html#Namespace_Names

include/Utils.h
```diff
@@ -31,6 +28,10 @@ StructType *unwrapValuePointerToStruct(Value *value);
 bool IRstructNameEqualsCstructName(std::string &structName,
                                    std::string &optLoadFileName);
 
+} // namespace Dataflow
+
+namespace Util {
```
kelloggm, line 33: "Util" is too vague of a name for a namespace, IMO. It's likely to collide with others if we ever want to link this project against another. (This applies to "dataflow" too, to be fair.)

Would you consider namespaces like "rlcdataflow" and "rlcutil"? I think those are a lot less likely to cause collisions in the future, but they're still clear.

src/Analysis.cpp
```diff
@@ -118,10 +118,11 @@ void buildCFG(CFG &topCFG, std::vector<std::string> branchOrder,
     for (auto succ : succs) {
       std::string succName = succ->getParent()->getName().str();
 
-      if (succName == branchName)
+      if (succName == branchName) {
         continue;
+      }
 
-      if (cfgMap.count(succName) > 0) {
+      if (cfgMap.count(succName)) {
```
kelloggm, line 125: I prefer the explicit "> 0" as a programmer who works primarily in not-C/C++, but if this is idiomatic the change is ok.

kelloggm: A brief internet search suggests that this is idiomatic in C++, so this change can stand.

src/Analysis.cpp
```diff
@@ -538,8 +540,6 @@ void CodeAnalyzer::doAnalysis(Function &F, std::string optLoadFileName) {
   std::map<std::string, InstructionHolder> branchInstructionMap;
 
   for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I) {
-    WorkSet.insert(&(*I));
```
kelloggm, line 541: This change is not semantics-preserving (I think). Is it intentional? If so, what's the motivation for including it in this PR that otherwise is just style changes?

src/Annotations/Annotation.cpp
```diff
@@ -441,16 +442,16 @@ bool rawStringIsCorrectlyFormatted(const std::string &rawAnnotationString) {
   return true;
 }
 
-Annotation *generateAnnotation(const std::string &rawAnnotationString) {
-  if (!rawStringIsCorrectlyFormatted(rawAnnotationString)) {
+Annotation *
+Annotation::generateAnnotation(const std::string &rawAnnotationString) {
```
kelloggm, line 446: I assume that the auto-formatter is responsible for this, but it looks bad - this probably shouldn't need to be on two lines. I'd consider renames here to get this back on one line. For example, the parameter could be called "rawAnnoString", "rawAnnoStr", or even just "rawAnno" (since its type is string)

kelloggm (changes requested):

src/Analysis.cpp
Kevin-TD: WorkSet is unused

kelloggm (approved):
LGTM

PR #20, opened 2023-12-13
Replacement of Clang Format

Kevin-TD:
astyle-format version 3.1 used in code and in ci


kelloggm (approved):
LGTM. I only looked at the changes to the CI/docker config and the python scripts, under the assumption that the changes in the C code are all style-only.

PR #21, opened 2023-12-27
Bug fixes for disjoint set refactor

Kevin-TD:
Bug 1: A struct is referenced in a different branch than the one it was defined in for certain gepInst instructions. This caused segfaults in the alias reasoning. Fixed by creating a function to search an entire program function for a certain value pointer or instruction.

Bug 2: addAlias method of DisjointedPVAliasSets did not handle case where both sets exist. This case is now explicitly handled. addAlias code also cleaner.

Additional test cases added

### DIFF
--- a/Testers/Passes/double_mustcall/index.txt
+++ b/Testers/Passes/double_mustcall/index.txt
@@ -1,0 +1,1 @@
+pass=CalledMethods function=main var=s.0 methods={free,free1}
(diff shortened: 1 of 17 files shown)


kelloggm:
@Kevin-TD sorry I did not get to review this quickly. This PR is not self-contained, and it should be split into two PRs: one for each of the bugs you mentioned in your description. Please open two near PRs that properly separate the two different bug fixes.

Also, it looks like CI is failing. Please fix those issues before you assign me to review the new PRs.

PR #22, opened 2024-01-09
Add alias bug fix for refactor branch

Kevin-TD:
Bug desc: addAlias method of DisjointedPVAliasSets did not handle case where both sets exist. This case is now explicitly handled. addAlias code also cleaner.


kelloggm (approved):
LGTM

PR #23, opened 2024-01-09
Bug fix regarding finding value ptr for disjoint set refactor branch

Kevin-TD:
Bug desc: A struct is referenced in a different branch than the one it was defined in for certain gepInst instructions. This caused segfaults in the alias reasoning. Fixed by creating a function to search an entire program function for a certain value pointer or instruction.

### DIFF
--- a/include/ProgramRepresentation/PVAliasSet.h
+++ b/include/ProgramRepresentation/PVAliasSet.h
@@ -26,9 +26,7 @@ class PVAliasSet {
     // LLVM docs: https://llvm.org/docs/LangRef.html#identifiers
     bool contains(const std::string& cleanedName);
 
-    // returns true iff there is a program var in this set with Value* value.
-    // this overload should only be used if you are looking for value that may
-    // not exist in the branch you're currently analyzing.
+    // returns true iff there is a program var in this set with Value* value
     bool contains(Value* value);
 
     // adds programVar to this set of program variables. programVar is checked to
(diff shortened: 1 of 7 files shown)


include/ProgramRepresentation/PVAliasSet.h
```diff
@@ -26,6 +26,10 @@ class PVAliasSet {
     // LLVM docs: https://llvm.org/docs/LangRef.html#identifiers
     bool contains(const std::string& cleanedName);
 
+    // returns true iff there is a program var in this set with Value* value.
+    // this overload should only be used if you are looking for value that may
+    // not exist in the branch you're currently analyzing.
```
kelloggm, line 31: What is the consequence if this overload is used at the wrong time?

kelloggm: (same comment applies to the similar text in `ProgramPoint.h`)

Kevin-TD: Well, there would actually be no consequence since it would be as effective as searching by `ProgramVariable`. I wrote "this should only be used ..." at the time to emphasize that this overload was for this one specific bug fix, but I think I will delete that part of the comment. The comment in [`ProgramFunction.h`](https://github.com/Kevin-TD/c-resource-leak-checker/blob/0ab713868045c1d10c2d5a73c01f170d90aadb1e/include/ProgramRepresentation/ProgramFunction.h#L41) for the `getPVASRefFromValue` method explains the situation anyhow

kelloggm (approved):

PR #25, opened 2024-01-26
Specify field name in test 

Kevin-TD:
- When writing passes, field name can be specified so we can type, e.g., `my_struct.x` instead of `my_struct.0`
- Implemented with an AST pass (`TestHelpers/struct_field_to_index_map_generator.py`) 
- Test passes now refer by field name 
- More annotation tests added


TestHelpers/struct_field_to_index_map_generator.py
```diff
@@ -0,0 +1,312 @@
...
+    def __init__(self, name: str):
+        self.name = name
+        self.typedefs: list[str] = []
+        self.fields: list[Field] = []
+
+    def get_name(self) -> str:
+        return self.name
+
+    def get_typedefs(self) -> "list[str]":
+        return self.typedefs
```
kelloggm, line 58: No `get_fields`? That's surprising.

TestHelpers/struct_field_to_index_map_generator.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+class StructVarManager:
+    def __init__(self):
+        self.struct_vars: list[StructVar] = []
+
+    # if there already exists a var of name var_name, that struct is returned.
+    # otherwise, the var is added and returned.
+    def add_struct_var(self, var_name: str, type_name: str) -> bool:
+        for struct_var in self.struct_vars:
```
kelloggm, line 122: When I run `pylint` on this file, I get a large number of [W0621 warnings](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/redefined-outer-name.html) about redefining variable names. These should probably all be fixed, since they can be nasty to debug. Here's the full list:
```
struct_field_to_index_map_generator.py:43:40: W0621: Redefining name 'field_index' from outer scope (line 236) (redefined-outer-name)
struct_field_to_index_map_generator.py:60:24: W0621: Redefining name 'field' from outer scope (line 307) (redefined-outer-name)
struct_field_to_index_map_generator.py:122:12: W0621: Redefining name 'struct_var' from outer scope (line 301) (redefined-outer-name)
struct_field_to_index_map_generator.py:131:33: W0621: Redefining name 'structs_manager' from outer scope (line 239) (redefined-outer-name)
struct_field_to_index_map_generator.py:138:31: W0621: Redefining name 'field_index' from outer scope (line 236) (redefined-outer-name)
struct_field_to_index_map_generator.py:152:4: W0621: Redefining name 'field' from outer scope (line 307) (redefined-outer-name)
struct_field_to_index_map_generator.py:156:34: W0621: Redefining name 'struct_var_manager' from outer scope (line 240) (redefined-outer-name)
struct_field_to_index_map_generator.py:156:72: W0621: Redefining name 'structs_manager' from outer scope (line 239) (redefined-outer-name)
struct_field_to_index_map_generator.py:193:22: W0621: Redefining name 'structs_manager' from outer scope (line 239) (redefined-outer-name)
```

kelloggm: (pylint also gives a bunch of warnings that aren't very helpful about style, but these ones are worth paying attention to)

TestHelpers/struct_field_to_index_map_generator.py
```diff
@@ -0,0 +1,312 @@
+#! python3 struct_field_to_index_map_generator.py <ast txt file> <output for map file>
```
kelloggm, line 1: In a separate PR, you should add an autoformatter for the python code in this repo. I suggest [black](https://github.com/psf/black), but any autoformatter will do. (For a discussion of the options, [this post](https://www.locallyoptimal.com/blog/2019/08/23/why-you-should-use-black-for-your-python-style-linting/) is very good. I have my students in CS 490 read this before our discussion of autoformatting, because the rationales it discusses are quite reasonable and reflect good engineering practice.)

TestHelpers/struct_field_to_index_map_generator.py
```diff
@@ -0,0 +1,312 @@
...
+
+        else:
+            if "VarDecl" == decl and "struct" in expr:
+                add_struct_var(expr[first_letter_index:],
+                               struct_var_manager, structs_manager)
+
+            field_index = None
+
+    with open(output_file, "w") as output:
+        for i in range(len(struct_var_manager.struct_vars)):
```
kelloggm, line 300: pylint suggests:
```
struct_field_to_index_map_generator.py:300:8: C0200: Consider using enumerate instead of iterating with range and len (consider-using-enumerate)
struct_field_to_index_map_generator.py:306:12: C0200: Consider using enumerate instead of iterating with range and len (consider-using-enumerate)
```

I agree.

TestHelpers/struct_field_to_index_map_generator.py
```diff
@@ -0,0 +1,312 @@
...
+
+class StructVarManager:
+    def __init__(self):
+        self.struct_vars: list[StructVar] = []
+
+    # if there already exists a var of name var_name, that struct is returned.
+    # otherwise, the var is added and returned.
+    def add_struct_var(self, var_name: str, type_name: str) -> bool:
+        for struct_var in self.struct_vars:
+            if struct_var.get_var_name() == var_name:
```
kelloggm, line 123: Rather than loop + if, could you use a `contains` method (or just `in`)? I think there's a more python-idiomatic ("pythonic"?) way to do this.

TestHelpers/struct_field_to_index_map_generator.py
```diff
@@ -0,0 +1,312 @@
...
+
+    field = Field(field_name, field_index)
+    struct.add_field(field)
+
+
+def add_struct_var(var_decl: str, struct_var_manager: StructVarManager, structs_manager: StructsManager):
+
+    var_decl_chunks = var_decl.split(" ")
+
+    print(var_decl_chunks)
```
kelloggm, line 160: Should probably be `logout` rather than `print`

TestHelpers/struct_field_to_index_map_generator.py
```diff
@@ -0,0 +1,312 @@
...
+        decl = ""
+        cur_decl_char = expr[start_decl_index]
+
+        try:
+            while (cur_decl_char != " "):
+                decl += cur_decl_char
+                start_decl_index += 1
+                cur_decl_char = expr[start_decl_index]
+        except IndexError:
+            # might be a null stement (looks like "<<<NULL>>>" in AST)
```
kelloggm, line 256: I'd prefer that we avoid exceptional control flow here. Can you check directly for `<<<NULL>>>` before you try to access the array instead of catching the `IndexError`?

(Exceptional control flow is both much slower than an `if` statement and tends to be harder for humans to reason about, in my experience.)

Testers/Annotations/double_mustcall/index.txt
```diff
@@ -0,0 +1,3 @@
+MustCall target = STRUCT(M).FIELD(0) methods = free, free1
```
kelloggm, line 1: As a general rule, anything that can be automatically generated shouldn't be checked into source control. This is also why e.g., we don't check in Java `.class` files or `.o` object files. 

These `.txt` files look automatically generated to me (in particular, the new ones in this PR, like this file, that reference fields by index). Can we avoid checking them in and instead automate their production?

kelloggm (changes requested):

TestHelpers/struct_field_to_index_map_generator.py
Kevin-TD: That was because I was defining setters and getters but they weren't actually being enforced so I could still just get `.fields`; fixing by making the members of `Struct` private by adding `__` to their name (e.g., `self.name` => `self.__name`)

TestHelpers/struct_field_to_index_map_generator.py
Kevin-TD: I have installed a formatter in the `test-suite-qol-additions` branch and have isolated the formatter into the `python-formatter` branch; it uses autopep8. I'll make the PR for it in a bit

TestHelpers/struct_field_to_index_map_generator.py
Kevin-TD: Since this method checks for equality by comparing two different types, `StructVar` and `str` (`var_name` ), I feel defining a contains method for this specific case would be awkward

TestHelpers/struct_field_to_index_map_generator.py
Kevin-TD: It's hard for me to directly check for null statements without making too many assumptions about the clang AST string. It's also hard to clear up my confusions about it since the documentation on it is very poor (as far as I've searched). Thus, this exception is kind of a catch-all for any unfamiliar AST statements; if there are more unfamiliar AST statements I come across, I'll try to update this section to be more clear on what types of strange statements we're looking for

Testers/Annotations/double_mustcall/index.txt
Kevin-TD: The files are not automatically generated and serve as the ground truth/reference for what the annotation generator should output. The annotation tester works by running the current annotation generator and running `diff` on its current output and the corresponding file in `Testers/Annotations`

TestHelpers/struct_field_to_index_map_generator.py
kelloggm: Even if we want to keep the exceptional control flow for unexpected statements, for AST statements that we **know** cause this flow (at least `<<<NULL>>>`, possibly others), it would be better to add an explicit check.

I'm okay with leaving it as-is if you add a TODO indicating that the documentation is too poor for you to do better.

kelloggm (approved):
I don't need to review again after my one comment is addressed.

PR #27, opened 2024-01-31
Adds temp file manager

Kevin-TD:
temp file manager is used to simplify the process of creating AST on the user end. we can use it to generate only one AST and pass it around as needed. here, we need the AST for `getAnnotationStrings` and for the `StructFieldToIndexMap`


kelloggm (approved):
LGTM, very nice improvement

PR #28, opened 2024-02-01
Pass & Annotation tests revamped 

Kevin-TD:
-Pass and Annotation test frameworks function as CLTs 
-- Pass test framework users to easily specify the disabling make call, disabling IR generation, disabling IR generation for specific files, not running specific files, only and running a specific file via short flags (e.g., --no-make, --no-build-ir-for main.c)
-- Annotation test framework allows users to specify files they do or don't want to test 
-Full tests on Pass and Annotation are now managed through `whole_test_run.py`

Implementation details
-`Flag.py`, `FlagsManager.py` represent user input flags (e.g., --no-make)
-`TestResult.py` keeps track of the test result (pass, failed, ignored) and any additional notes we may have (e.g., the exit status code if it failed) 

-`TestFile.py`, `TestFilesManager.py` are base classes that represent the test files in the `/test` directory. These classes are extended and modified for each type of test (Pass and Annotation) 
-`TestRunner.py`is an abstract class that executes the test and must be extended/implemented for each type of test  
-- The extensions of these base classes are in `TestRunners/AnnotationTestRunner` and `TestRunners/PassTestRunner`


TestRunners/TestRunner.py
```diff
@@ -0,0 +1,181 @@
...
+                raise ValueError(
+                    f"Flag {arg} not found. {self._flags_manager.to_string()}")
+
+            sys_arg_iterator += 1
+
+        self._before_tests_run()
+
+        for c_file in c_files:
+            f = self._test_files_manager.get_file(
+                c_file[find_nth(c_file, "/", 3) + 1:])
```
kelloggm, line 156: Please add a comment explaining the constant indices used in this string operation.

TestRunners/TestResult.py
```diff
@@ -0,0 +1,81 @@
...
+        self.__object_type = "Folder"
+
+    def set_object_type_to_file(self):
+        """"When to_string is called, test will be referred to as a file. 
+        This is the default behavior and does not need to be called if the test 
+        is intended to represent a file and not a folder. 
+        """
+        self.__object_type = "File"
+
+    def test_has_passed(self, note=""):
```
kelloggm, line 37: I find the naming convention you've chosen in this file confusing, because `test_has_failed` sounds like a bool.

I suggest the following names:
* `has_passed()`/`has_failed()` for the booleans (i.e., the methods that now start with `check...`
* `pass()`/`fail()` for the void operations that set the test output status (this matches the standard xUnit terminology, as a bonus)

TestRunners/TestResult.py
```diff
@@ -0,0 +1,81 @@
...
+    def test_has_failed(self, note=""):
+        if note != "":
+            self.add_note(note)
+
+        self.__test_status = self.__TEST_STATUS["FAILED"]
+
+    def check_if_test_has_failed(self) -> bool:
+        return self.__test_status == self.__TEST_STATUS["FAILED"]
+
+    def test_is_ignored(self, note=""):
```
kelloggm, line 52: I'd also change the name of this method to `ignore()`

TestRunners/Utils.py
```diff
@@ -0,0 +1,73 @@
...
+TERMINAL_TEXT_STYLES = {
+    "BOLD_ON": "\033[1m",
+    "BOLD_OFF": "\033[22m",
+    "RED_COLOR": "\033[31m",
+    "GREEN_COLOR": "\033[32m",
+    "RESET_COLOR": "\033[0m",
+}
+
+
+def split_string_by_n_characters(input_string: str, n: int) -> "list[str]":
```
kelloggm, line 12: I suggest writing unit tests for each of the methods in this file.

kelloggm: (and running them in CI)

run_at.py
```diff
@@ -0,0 +1,16 @@
+import sys
+import os
+
+# Alias for ../TestRunners/AnnotationTestRunner/run_test.py
```
kelloggm, line 4: Why is this necessary?

run_pt.py
```diff
@@ -0,0 +1,16 @@
+import sys
+import os
+
+# Alias for ../TestRunners/PassTestRunners/run_test.py
```
kelloggm, line 4: Why is this necessary?

test/double_mustcall/index.c
```diff
@@ -0,0 +1,20 @@
...
+#include <stdlib.h>
+#include <unistd.h>
+#include "../../Annotations/Annotations.h"
+
+struct M {
+    char* x MustCall("free, free1");
+    char* y MustCall("free");
+};
+
+void free1(void* p Calls("free1")) {
```
kelloggm, line 11: Not really related to this PR, but I think I just realized that I don't understand the semantics of your `Calls` annotation. I would have expected the annotation here to be `Calls("free")` instead of `Calls("free1")`, since this method calls `free()` on `p`; this method _is_ `free1`, so it doesn't make sense to write that it calls itself on its own argument (which seems to me to be a tautology). We should discuss this (maybe in our next meeting or over email).

test/if_lub_gepinst_test/index.c
```diff
@@ -0,0 +1,35 @@
...
+    free(p);
+}
+
+int main() {
+    struct M s;
+
+    s.x = (char*)malloc(15);
+
+    s.y = s.x; // note: when we alias s.y and s.x, the mc of the set in total will be {"free", "free1"}
+    // how does this happen? we *first* do the alias reasoning and find out s.x <-> s.y,
```
kelloggm, line 21: Is the alias reasoning correctly combining the must-call sets? In the Java implementation, we wouldn't even permit this assignment (because x's `MustCall(free, free1)` is not a subtype of y's `MustCall(free1)` (assignments in the other direction would be permitted)). With that context, I think the behavior here is correct: if we assign `x` to `y`, the safe thing to do is to take the worst case scenario (we have to fulfill `x`'s must-call obligation on `y`)

kelloggm: We may also want to discuss this in detail.

TestRunners/TestFile.py
```diff
@@ -0,0 +1,46 @@
+from .Utils import *
+
+
+class TestFile:
+    """ represents a file to test with information on whether certain properties will be ran.
```
kelloggm, line 5: I'm not sure how to parse this, especially "will be ran". I think you mean "represents a file to test, along with information about how to run the test", but I'm not really sure.

TestRunners/PassTestRunner/run_test.py
```diff
@@ -0,0 +1,77 @@
...
+    flags_manager.add_flag(
+        "f", "no-build-ir-for",
+        "Does not build IR for the specified c file. Can be called multiple times. c file must be specified in the succeeding argument.",
+        lambda pass_test_files_manager, file_name: pass_test_files_manager.get_file(
+            file_name).toggle_ir_generation()
+
+    )
+    flags_manager.add_flag(
+        "u", "only-build-ir-for",
+        "Only builds ir for specified c file. c file must be specified in the succeeding argument. Note that c files will still be ran unless otherwise disallowed (use -n). Combined with --no-build-ir may result in undesirable output as these commands are opposites.",
```
kelloggm, line 55: typo "still be ran" should be "still be run", I think (run is an irregular verb though so I might actually be wrong about this - English class!)

kelloggm: I'd prefer that combining with `--no-build-ir` leads to an exception if its non-sensical (in both the handling of this command and that one)

TestRunners/PassTestRunner/run_test.py
```diff
@@ -0,0 +1,77 @@
...
+        lambda pass_test_files_manager, file_name: [file.toggle_ir_generation(
+        ) for file in pass_test_files_manager.get_all_files_excluding(file_name)]
+    )
+
+    flags_manager.add_flag(
+        "r", "no-run-for",
+        "Does not specified c file. Can be called multiple times. c file must be specified in the succeeding argument. Combined with --only-run-for may result in undesirable output.",
```
kelloggm, line 62: typo "Does not specified c file" is missing a word (probably "run" or "test")

kelloggm: Also explain why `Combined with --only-run-for may result in undesirable output` and possibly throw an exception, depending on how bad it is.

TestRunners/AnnotationTestRunner/run_test.py
```diff
@@ -0,0 +1,51 @@
...
+    flags_manager = FlagsManager(
+        "python3 ../run_at.py [required test folder name] [optional flags]")
+    flags_manager.add_flag(
+        "h", "help",
+        "Show help"
+    )
+
+    flags_manager.add_flag(
+        "r", "no-run-for",
+        "Does not check annotations for specified c file. Can be called multiple times. c file must be specified in the succeeding argument. Combined with --only-run-for may result in undesirable output.",
```
kelloggm, line 35: Should we throw an exception here if both flags are passed?

kelloggm (changes requested):
This review was a bit superficial on some of the files, because I'm not a Python expert and I got tired about halfway through (sorry - it's a lot of code). I'll try to focus on the other files that I didn't look at closely the next time that I review, but many of my comments apply throughout.

TestRunners/TestResult.py
Kevin-TD: I wish I could use `pass` but that's a special keyword in Python, so I'll set it to `pass_test` and `fail_test` + changing bool methods to `has_passed` and `has_failed`

TestRunners/Utils.py
Kevin-TD: Considering how small each function is and how their importance on the codebase is relatively small, what usage is obtained by writing unit tests for them? I feel all the testing I did while developing the Python code made me pretty confident in the reliability of these functions

run_at.py
Kevin-TD: Purely convenience. Same reason for `run_pt.py`; easier to remember and type out `python ../run_at.py testname` than `python ../TestRunners/AnnotationTestRunner/run_test.py testname`, though I can always `.gitignore` them if you feel this does not justify their existence

test/double_mustcall/index.c
Kevin-TD: This is a great point! Logic fix coming soon in a new pr. I agree, it shouldn't make sense that the code works this way. Its purpose here is to try to say that `p` calls `free1`, but the code should (and is definitely able to) pick that up automatically whenever we actually call `free1`. `Call("free")` would be the appropriate annotation on `p`.

TestRunners/Utils.py
kelloggm: > I feel all the testing I did while developing the Python code made me pretty confident in the reliability of these functions

As far as I'm concerned, this testing did not occur (because it is not automated, so I can't observe it). More importantly, going forward there is no way to guarantee that these functions will continue to work as expected without writing tests that cover them.

A good software engineering practice is to always write automated tests for any code that you write, check those tests in, and run them in CI, because manual testing is notoriously unreliable and doesn't preserve any record of the functions expected behavior.

run_at.py
kelloggm: That's reasonable. Please add this logic to the comment at the start of this file.

test/if_lub_gepinst_test/index.c
Kevin-TD: The assignment is improperly allowed here since annotation verification is not yet implemented; I did not catch the error myself. I'll update the test to be logically consistent

TestRunners/TestFile.py
Kevin-TD: Shortening to "represents a file to test." When I said "with information on whether certain properties will be ran" I was referring to properties like `__INIT_TEST_CONDITIONS.TEST_WILL_RUN` in `TestFile.py` and `__INIT_PASS_TEST_CONDITIONS.IR_WILL_GENERATE` in `PassTestFile.py`, but I think it's sufficient to just say it represents a file to test

TestRunners/PassTestRunner/run_test.py
Kevin-TD: Grammarly agrees with you that it'd be "run". But I've been using the phrase "undesirable output" to say that, because of how the CLT's are implemented, combining "opposite" flags may run the code exactly opposite to how the user expects

Foe example, let's say we call the test with flags `--no-run-for file1.c --only-run-for file1.c`, and say our test directory has files file1.c, file2.c, and file3.c. 
Each test file has an initial condition `TEST_WILL_RUN` set to true (defined in `TestFile.py`). The only way we can change it is by calling `TestFile.toggle_test_running`
As the flags are parsed from left to right, the following is executed:
`--no-run-for file1.c` => calls file1.toggle_test_running(), which now equals False
`--only-run-for file1.c` => calls toggle_test.running() for all files except `file1.c`. 
This results in no test running. 

I'll make it so an exception is thrown if "opposites" are called

kelloggm (approved):
LGTM now

PR #29, opened 2024-02-04
Adds and applies branch lister

Kevin-TD:
-branch lister added as a debug tool
-getSuccessor and getPredecessor changed to be a part of rlc_dataflow


kelloggm (approved):
LGTM

PR #35, opened 2024-02-11
Logic Fixes

Kevin-TD:
-removes allocation assumption (i.e., when a struct w/ MC annotations is initialized, it no longer assigns its fields internally MC's automatically) 
-tests changed to not set the calls annotation to itself 
-CM pass now adds the invoker function to the set of called methods, and not just the methods inside some annotation 
-tests updated to reflect called methods change


kelloggm:
@Kevin-TD can you resolve the conflict on this branch as soon as you get a chance? I'll then review this PR tomorrow.

include/CalledMethods.h
```diff
@@ -6,12 +6,13 @@
...
-    void onReallocFunctionCall(PVAliasSet *input, std::string &fnName);
-    void onSafeFunctionCall(PVAliasSet *input, std::string &fnName);
-    void onAnnotation(PVAliasSet *input, std::string &fnName,
+    void onAllocationFunctionCall(PVAliasSet* input, std::string &fnName);
+    void onDeallocationFunctionCall(PVAliasSet* input, std::string &fnName);
+    void onUnknownFunctionCall(PVAliasSet* input);
+    void onReallocFunctionCall(PVAliasSet* input, std::string &fnName);
+    void onSafeFunctionCall(PVAliasSet* input, std::string &fnName);
+    void onAnnotation(PVAliasSet* input, std::string &fnName, AnnotationType annotationType);
+    void onAnnotation(PVAliasSet* input, std::string &annoFnName, std::string& invokerFnName,
```
kelloggm, line 15: When there are two methods with the same name but different parameter lists in a header file, it's a good idea to include a comment indicating the difference between the two. Can you do that here? The goal is that someone reading this file should be able to figure out which of the two is the right one to use in a given situation without needing to read the .cpp file.

include/DataflowPass.h
```diff
@@ -20,6 +20,7 @@ class DataflowPass {
                     std::string priorBranch);
     void transfer(Instruction *instruction, ProgramPoint &inputProgramPoint);
     void insertAnnotation(Annotation *annotation, PVAliasSet *pvas);
+    void insertAnnotation(Annotation *annotation, PVAliasSet *pvas, std::string& invokerFnName);
```
kelloggm, line 23: Same comment here as in `CalledMethods.h`

include/DataflowPass.h
```diff
@@ -71,7 +72,7 @@ class DataflowPass {
 
     // a helper function that checks for parameter annotations on call
     // instructions. returns true if an annotation was found, and false if not.
-    bool handleIfAnnotationExistsForCallInsts(const std::string &fnName,
+    bool handleIfAnnotationExistsForCallInsts(std::string &fnName,
```
kelloggm, line 75: Why is this no longer `const`? The comment here doesn't give any indication that this method might change any of its parameters. If it might change `fnName`, the comment above needs to indicate that.

include/DataflowPass.h
```diff
@@ -91,9 +92,23 @@ class DataflowPass {
...
+    // annoFnName is the name of the function inside the annotation
+    // invokerFnName is the name of the function the annotation belongs to
+    /*
+    e.g.,
+    void free0(void* p Calls("free")) {
+      free(p);
+    }
+
+    here, annoFnName = "free" and invokerFnName = "free0"
+    */
```
kelloggm, line 108: This comment is very helpful, thanks.

src/CalledMethods.cpp
```diff
@@ -26,9 +26,16 @@ void CalledMethods::leastUpperBound(PVAliasSet &preSet, MethodsSet &curMethodsSe
     preSet.methodsSetIntersection(curMethodsSet);
 }
 
-void CalledMethods::onAnnotation(PVAliasSet *input, std::string &fnName,
-                                 AnnotationType annotationType) {
+void CalledMethods::onAnnotation(PVAliasSet *input, std::string &fnName, AnnotationType annotationType) {
```
kelloggm, line 29: Why do we need this version of this method? Shouldn't it be the case that we automatically infer a `Calls` "annotation" (internally, without necessarily representing it as an annotation) whenever we see a method call, even if the parameter is unannotated. For example, consider the following:
```
void free(void* ptr) { ... }
free(p);
```

After the call to `free(p)`, the Called Methods pass should record that `free` has been called on `p`. No annotation should be needed to express this: the pass should handle it automatically, for all method calls. In particular, there should be no need for a user to write something like:
```
void free(Calls("free") void* ptr) { ... }
```

The existence of this method vs the method below indicates that something about how this should work is not quite right: in particular, we should only need one method to handle annotations, and it should only add the methods that are actually defined in the annotation's body (I think that's `annoFnName` here, if I understand correctly). So, another way of saying this is that I don't understand the need for the other `onAnnotation` method: for all annotated parameters, we should be including the invoker's name automatically (since we should be doing that for all parameters, not just those that are annotated).

kelloggm (changes requested):

include/DataflowPass.h
Kevin-TD: `const` initially removed out of low desire to annotate all the other places where `const` would need to be added to allow this function to be annotated with `const`. I'll be adding back to `const` here + all other places where `const` is needed for this function to be `const`

include/DataflowPass.h
```diff
@@ -106,7 +107,72 @@ class DataflowPass {
...
+    }
+    -----------
+
+    This code finds that there is a ReturnAnnotation on the function
+    free_and_obligate and will add "free" to return.x's called methods. Here, B is
+    set to that return value so B.x will have called methods {"free"}. But if we
+    also say to add the invoker name to the called methods (free_and_obligate),
+    then B.x's called methods will be {"free", "free_and_obligate"}. This is not
+    correct, hence the need to be able to include or and not include
+    the invoker name.
```
kelloggm, line 145: I would expect that return annotations and the standard rules for adding dataflow facts do not interact in cases like this. In particular, I'd expect that the code handles the following entirely separately:
* process the fact that `free_and_obligate` is called on `A`. If we happen to also have some other fact that allows us to conclude that B and A are aliases, this fact can also be propagated to B. (In the Java RLC, this would be handled by the Returns Receiver checker, which would provide this kind of fact.)
* process the return annotation, which only applies to the `.x` field of the returned value; this should be propagated to `B.x` (and then `A.x` from there, again, iff A and B are known to be aliases).

The resulting dataflow store after this call thus looks like (assuming A and B are in fact aliases):
```
A = B -> CalledMethods("free_and_obligate")
A.x = B.x -> CalledMethods("free")
```

If A and B were not (known to be) aliases, we'd instead have:
```
A -> CalledMethods("free_and_obligate")
B -> no information/top
A.x -> no information/top
B.x -> CalledMethods("free")
```

include/DataflowPass.h
```diff
@@ -106,7 +107,72 @@ class DataflowPass {
...
+    }
+
+    int main() {
+      char* a = (char*)malloc(15);
+      does_free(a);
+    }
+    -----------
+
+    Here, it is appropriate that the called methods
+    of a = {"free", "does_free"}
```
kelloggm, line 172: Agreed, but I'd expect this to happen due to two different calls (one for the annotation, and one for the dataflow fact) that are then GLB'd.

kelloggm (changes requested):

Testers/Passes/includes_test/call_structs.txt
```diff
@@ -4,7 +4,7 @@ pass=MustCall function=main var=m.y methods={free}
 pass=MustCall function=main var=km.x methods={free}
 pass=MustCall function=main var=y methods={free}
 pass=CalledMethods function=main var=random_var methods={free}
-pass=CalledMethods function=main var=k.x methods={free}
-pass=CalledMethods function=main var=k.y methods={free}
-pass=CalledMethods function=main var=m.y methods={free}
-pass=CalledMethods function=main var=km.x methods={free}
\ No newline at end of file
+pass=CalledMethods function=main var=k.x methods={creates_obligation,free}
+pass=CalledMethods function=main var=k.y methods={creates_obligation,does_free,free}
```
kelloggm, line 8: I don't understand why either `k.x` or `k.y` should have `creates_obligation` in their CM set here. That method is only called on `k`, not on its fields.

src/CalledMethods.cpp
```diff
@@ -26,9 +27,17 @@ void CalledMethods::leastUpperBound(PVAliasSet &preSet, MethodsSet &curMethodsSe
...
+void CalledMethods::onAnnotation(PVAliasSet* input, Annotation* annotation, const std::string& invokerFnName) {
+    if (annotation->getAnnotationType() == AnnotationType::CallsAnnotation) {
+        auto annoMethods = annotation->getAnnotationMethods();
+        for (std::string annoMethod : annoMethods) {
+            input->addMethod(annoMethod);
+
+        }
+
+        if (ParameterAnnotation *paramAnno = dynamic_cast<ParameterAnnotation *>(annotation)) {
+            input->addMethod(invokerFnName);
```
kelloggm, line 39: I don't think this is correct: it doesn't match my intuition. This might be why I'm confused about the expected output of a test case, too. Let me explain my thinking:
* when some method is called, e.g., `foo(a, b)`, regardless of the annotations on the parameters of `foo`, I expected `a` and  `b` to have `foo` in their CM sets. This should happen regardless of annotations.
* if there are annotations on the parameters (e.g., suppose the second parameter of `foo` has `Calls("free", "x")`, I expect that those are _also_ applied, but that the two things should happen separately. So, after this call to foo, I'd expect the CM store to look like: `{ a -> CM("foo"), b -> CM("foo"), a.x -> CM("free") }`. However, I think this code results in the following CM store: `{ a -> CM("foo"), b -> CM("foo"), a.x -> CM("free", "foo") }`. Note the extra "foo" in `a.x`'s CM set, which is what I'm concerned about.

Testers/Passes/includes_test/call_structs.txt
Kevin-TD: I have it like this because of an implementation detail I should've mentioned sooner; since the IR destructures fields into its elements, internally the called methods/must call set of `k` (referring to the struct in this test case) does not hold any useful information about the resources we need to track. `k` is just a wrapper referring to its fields which actually contain the resources, hence why we treat something like `creates_obligation(k)` as `creates_obligation(k.x, k.y)`

kelloggm: This worries me. What if the call was to `free(k)`? If `k.x` or `k.y` had been separately allocated, this would lead to a false negative - so this is a serious problem. For example, consider the following example (from https://stackoverflow.com/questions/13590812/c-freeing-structs):
```
typedef struct Person
{
char * firstname , *last surName;
}Person;
Person *ptrobj =malloc(sizeof(Person)); // memory allocation for struct
ptrobj->firstname = malloc(n); // memory allocation for firstname
ptrobj->surName = malloc(m); // memory allocation for surName

.
. // do whatever you want

free(ptrobj->surName);
free(ptrobj->firstname);
free(ptrobj);
```

If either of the calls to `free(ptrobj->surName)` or to `free(ptrobj->firstname)`, my understanding from what you're saying is that we'd still put `free` in the called methods sets of the fields.

Kevin-TD:
Review requested but not yet for immediate merging as one of the tests fails due to Major Issue*. Please review logical updates made
PR will be made to resolve Major Issue. Changes made (copy-pasted from the description on commit [8222ea1](https://github.com/Kevin-TD/c-resource-leak-checker/pull/35/commits/8222ea10f181ccd7559bd53a5da7608733a41508)) 

-'onFunctionCall' method added to passes; used simply state when an arg has been passed into a method
-Some passes remove mention of function name/invoker name 
-bug fix with sret call (lines 344-352 of DataflowPass.cpp) 
-method made specifically for handling structs with field annotations 
(DataflowPass::handleIfStructTyAndIfFieldsHaveAnnotations) 
-more convenience files run_wa and run_wp 

*Major Issue: IR does de-sugaring and we cannot reverse it. Revamping of AST passes will be made to 1) resolve the issue and 2) make all AST passes cleaner and easier to read

kelloggm (approved):
Sorry for the very long review cycle @Kevin-TD - I was traveling last week and didn't have time to page all of this code back in.

Your latest set of changes looks reasonable to me, and resolves my concerns (excepting the known bug that you mentioned). I'd prefer that that bug be addressed in a separate PR, so I'm okay with merging this as-is.

PR #36, opened 2024-02-29
Bug fix: detecting multiple annotations on one parameter

Kevin-TD:
Bug description: On structs with a small number of fields (typically 1 or 2), the LLVM IR destructures them into their elements. This conflicts with how parameters do not do this destructuring, making it so we would miss out on annotations on parameters with more than 1 annotation

Bug fix: iterate to find all possible parameter specifications, instead of just a singular index (`argIndex`)

The branch name specifies "small struct" because the IR does not destructure it when there many fields to the struct, causing it to pass it in with a byval attribute (see docs: https://llvm.org/docs/LangRef.html#parameter-attributes) or sret attribute (handled with DataflowPass::handleSretCallForCallInsts) 

This code does not fix it when it's passed in with byval, but the fix will be implemented in branch `multi-anno-one-param-fix`

### DIFF
--- a/Testers/Passes/call_all_small_test/call_2.txt
+++ b/Testers/Passes/call_all_small_test/call_2.txt
@@ -1,0 +1,4 @@
+pass=MustCall function=main var=test.a methods={free}
+pass=MustCall function=main var=test.b methods={free}
+pass=CalledMethods function=main var=test.a methods={calls_all,free}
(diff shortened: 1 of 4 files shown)


kelloggm:
@Kevin-TD Please resolve the conflict, then you can merge this.

PR #38, opened 2024-03-06
Bug fix on certain field type

Kevin-TD:
Field type that were declared using typedef alias were being parsed incorrectly by field-to-index map. Now fixed 

Note: it is known we need a testing framework for the struct field to index map to prove something was actually fixed. But consider that this is a pr for a bug fix for a pr for a bug fix.


kelloggm (approved):
> consider that this is a pr for a bug fix for a pr for a bug fix

Ok, LGTM

PR #39, opened 2024-04-04
Review AST Info Generator 

Kevin-TD:


testAST.txt
kelloggm, line 1: This file probably shouldn't have been committed (maybe add it to the .gitignore?)

ASTInfoGenerator/ast_info_generator.py
```diff
@@ -0,0 +1,190 @@
...
+            struct_var_manager.add_struct_var(
+                type_parsed.get_struct_var_name(), type_parsed.get_struct_type_name())
+
+        elif found_type is AnnotateAttr:
+            anno = Annotation(type_parsed.get_annotation_type(
+            ), type_parsed.get_annotation_target(), type_parsed.get_annotation_methods())
+            annotation_manager.add_annotation(anno)
+
+    if DEBUG:
+        for spec in specifier_manager.get_specifiers():
```
kelloggm, line 94: This debug printing code shouldn't be separate from the code that writes the result just below - if they ever diverge, it'll be extremely inconvenient to debug. Instead, structure the code in the following way:
* write a helper function that produces a (large) string
* make this code call that helper and then print the string
* make the output-writing code below call the same helper function and then write the string to a file

ASTInfoGenerator/DeclParser/DeclParser.py
```diff
@@ -0,0 +1,382 @@
...
+
+class DeclParser:
+    def __init__(self):
+        self.__param_index = None
+        self.__field_index = None
+
+    def _parse_function_decl(self, function_decl: str) -> Specifier:
+        # Assumption: functions that are extern are not relevant to our analysis
+        if function_decl.endswith("extern"):
+            return None
```
kelloggm, line 22: Why? `extern` doesn't really mean anything on functions - isn't external linkage the default?

kelloggm: My understanding of `extern` might be straight out of 1999 though

ASTInfoGenerator/DeclParser/DeclParser.py
```diff
@@ -0,0 +1,382 @@
...
+            f"function name = '{function_name}', return type = '{return_type}'")
+        return FunctionDecl(function_name, return_type)
+
+    def _parse_record_decl(self, record_decl: str):
+        record_decl_chunks = record_decl.split(" ")
+        struct_name = record_decl_chunks[len(record_decl_chunks) - 2]
+
+        return RecordDecl(struct_name)
+
+    def _parse_parm_var_decl(self, parm_var_decl: str, param_index: int):
```
kelloggm, line 78: Nit: the inconsistency between uses of `parm` and `param` here bothers me a bit. I would find it easier to read this code if we use `param` everywhere.

ASTInfoGenerator/DeclParser/DeclParser.py
```diff
@@ -0,0 +1,382 @@
...
+        return FieldDecl(field_name, field_index)
+
+    # only checks for vars whose type is struct
+    def _parse_var_decl(self, var_decl: str, specifier_manager: SpecifierManager):
+        var_decl_chunks = var_decl.split(" ")
+
+        logout(var_decl)
+        logout(var_decl_chunks)
+
+        if var_decl_chunks[5] == "used":
```
kelloggm, line 153: There is some kind of assumption being made here that I don't understand. What's the significance of "used"? I think a comment is needed here to explain.

ASTInfoGenerator/DeclParser/DeclParser.py
```diff
@@ -0,0 +1,382 @@
...
+                f"did not find for parsed var_decl var name = {var_name} struct name = '{struct_name}'")
+            return None
+
+        logout(var_decl)
+        logout(
+            f"did parsed var_decl var name = {var_name} struct name = '{struct_name}'")
+
+        return StructVarDecl(var_name, struct_name)
+
+    def _parse_anno(self, anno_decl: str, specifier_manager: SpecifierManager, param_index: int, field_index: int):
```
kelloggm, line 195: Don't we already have annotation parsing code in Python in one of your other helper files? If so, would it be possible to combine this parser and that one? If not, why not?

ASTInfoGenerator/DeclParser/DeclParser.py
```diff
@@ -0,0 +1,382 @@
...
+                    # Assumption: spec.return_type looks like 'struct <struct_name> ' or '<struct_name> '
+                    # note: it'll look like '<struct_name> ' if a typedef is used
+
+                    # we can make this assumption because:
+                    # 1. param_index is none, which means we are making an
+                    # annotation for the function itself, and (for now) we assume
+                    # that means it's an annotation on its return.
+                    # note: we'll need to update how we treat annotations
+                    # applied to the function itself
+                    # if/when annotations like @Owning and @MustCallAlias are implemented
```
kelloggm, line 254: Will we? Both of these examples can be viewed as annotating the return type, too.

The only case that would be difficult is annotations that are explicitly about the function and not about the return; I don't think the RLC uses any of those besides `@EnsuresCalledMethods`.

ASTInfoGenerator/DeclParser/DeclParser.py
kelloggm: (Even if only some of the code can be shared, it would be worthwhile to share it.)

ASTInfoGenerator/DeclParser/DeclParser.py
```diff
@@ -0,0 +1,382 @@
...
+        decl_name = ""
+        cur_decl_char = expr[start_decl_index]
+
+        try:
+            while (cur_decl_char != " "):
+                decl_name += cur_decl_char
+                start_decl_index += 1
+                cur_decl_char = expr[start_decl_index]
+        except IndexError:
+            # might be a null statement (looks like "<<<NULL>>>" in AST)
```
kelloggm, line 341: This pattern appears at least one other place (`ast_info_generator.py`, around line 40). Can you write a helper function to handle this situation, so that if you do actually fix the `<<<NULL>>>` parsing issue, you only have to do so in one place?

kelloggm: As a general rule, copying a TODO statement is an anti-pattern: write a helper function instead so that you only have one copy of the TODO.

ASTInfoGenerator/Specifiers/SpecifierManager.py
```diff
@@ -0,0 +1,72 @@
...
+
+        if error_if_not_found:
+            raise ValueError(f"struct {struct_name} not found")
+
+        new_struct = Struct(struct_name)
+        self.__specifiers.append(new_struct)
+        return self.__specifiers[-1]
+
+    # if there already exists a function of name function_name, that function is returned.
+    # otherwise, the function is added and returned.
```
kelloggm, line 40: There doesn't seem to be much of a difference between `get_function` and `add_function` (and the same is true for the struct equivalents): get adds the function if it isn't found, and add gets the function if it is found. Can you combine the implementations?

testAST.txt
Kevin-TD: Yeah this file should not exist; deleting.

ASTInfoGenerator/DeclParser/DeclParser.py
Kevin-TD: I cannot recall why I did this in the past and removing those LoC did not change much output, so I'll be removing it.

ASTInfoGenerator/DeclParser/DeclParser.py
Kevin-TD: I will increase the usage of `param` but will keep `parm` in some places since the AST defines parameter declarations as `ParmVarDecl`s.

ASTInfoGenerator/DeclParser/DeclParser.py
Kevin-TD: There isn't much to explain other than that (what I've observed) for VarDecl's, the variable name follows the "used" keyword. How should I explain this in a comment?

ASTInfoGenerator/DeclParser/DeclParser.py
Kevin-TD: They will be combined in future where we make individual AST passes. The code I think you're referring to is `Annotations/annotation_generator.py`, and I intend on replacing it with a pass that just reads in and outputs the annotations generated by this info generator.

ASTInfoGenerator/DeclParser/DeclParser.py
Kevin-TD: I feel like we had a discussion that led to me writing the comment like this, but I cannot figure out the original rationale. Reading the code as well, I would agree it doesn't look like anything changes for those cases.

ASTInfoGenerator/DeclParser/DeclParser.py
Kevin-TD: Nevermind, I have found this definitely makes a difference. Way more functions are included in the final output, but I have no reason to not include it because it has an "extern." Many of these functions with "extern" are from standard C include files that do not concern our analysis, but it's very well possible a user writes a function with extern that we _should_ consider

ASTInfoGenerator/Specifiers/SpecifierManager.py
Kevin-TD: `get_function` was actually unused; deleting. I'll combine `get_struct` and `add_struct` into `get_or_add_struct`. The feature to error if not found was never actually used, so that parameter will be removed.

ASTInfoGenerator/DeclParser/DeclParser.py
kelloggm: I think a comment that shows the structure of a VarDecl here would make this clear, yes.

ASTInfoGenerator/DeclParser/DeclParser.py
kelloggm: Let's remove this comment, then.

ASTInfoGenerator/DeclParser/DeclParser.py
Kevin-TD: We can create a list but we don't know if that function came from what the user defined or if it's from the standard library. The AST does not preserve `#include` tags

kelloggm: Is it really a problem if we include functions from the standard library in the output? I agree they're clutter, but it's not _wrong_, correct?

Kevin-TD: Correct, it's not wrong to. But in the future I'll consider looking into seeing if I could filter them out, considering it changes AST info output from ~100 lines to 1000 lines

PR #40, opened 2024-04-13
Ast info reader

Kevin-TD:
- Creates reader for AST info 
- Makes passes for annotation generator + struct field to index map
- C++ code now passes around AST info instead of raw AST


ASTAnalyses/ASTInfo/generator.py
```diff
@@ -0,0 +1,159 @@
...
+from ASTAnalyses.ASTInfo.StructVariables.StructVarManager import *
+from ASTAnalyses.ASTInfo.Debug import *
+
+from ASTAnalyses.ASTInfo.DeclParser.DeclParser import *
+
+file_to_read = sys.argv[1]
+output_file = sys.argv[2]
+
+
+def is_null_stmt(line_of_ast: str):
```
kelloggm, line 21: Duplicated with `DeclParser.py` around line 350. Please consolidate these so that they share a single implementation.

src/Analysis.cpp
```diff
@@ -452,10 +452,12 @@ void CodeAnalyzer::doAnalysis(Function &F, std::string optLoadFileName) {
...
 
-        TempFileManager astTempFile("astTempFile");
-        tempfile_util::dumpASTIntoTempFile(optLoadFileName, astTempFile);
+        TempFileManager astInfoTempFile("astInfoTempFile");
+        tempfile_util::dumpASTInfoIntoTempFile(optLoadFileName, astInfoTempFile);
 
-        auto annotations = getAnnotationStrings(astTempFile);
+        TempFileManager annotationsTempFile = TempFileManager("annotationsTempFile");
+
+        auto annotations = getAnnotationStrings(astInfoTempFile); // TODO! FIX THIS
```
kelloggm, line 460: What's wrong with this? The "TODO! FIX THIS" seems bad - but I don't see the issue looking at this locally.

kelloggm (changes requested):
Am I right that `ASTAnalyses/ASTInfo/reader.py` and `ASTAnalyses/ASTInfo/generator.py` are inverses? That is, that `reader.py` is supposed to read in a file produced by `generator.py`? If so, that opens up a few opportunities that we should take advantage of:
* over a large corpus of C ASTs, we can write a test that checks that `reader.py` can always read in the output of `generator.py`, without needing to write any oracles (any case that `generator.py` produces something that `reader.py` can't read is always an error)
* we can make them share more code - for example, all of the strings that they use can be stored in one place and then used differently to read and to write, rather than be duplicated across both files. That will reduce the amount of places that need to change when we make modifications, and increase the chances that they exactly align.

src/Analysis.cpp
Kevin-TD: Comments not supposed to be there; deleting. It's all fine.

ASTAnalyses/ASTInfo/generator.py
Kevin-TD: It is unused in `generator.py` so I'll simply delete it from there

kelloggm (commented):
The set of changes since I last reviewed are a big improvement, and are exactly the kind of thing I was suggesting. This will make it much less likely that we have a mismatch between these implementations in the future.

CI is still failing, and I won't approve until it passes.

kelloggm (approved):

PR #43, opened 2024-08-07
Prevents duplicate parameter from appearing

Kevin-TD:
It was possible that a function's parameters could repeat twice. E.g., for function `char* do_thing(a, b, c)`, after it's decompressed by the AST and read again by our analysis, the function declaration for it may appear several times, alongside its parameter declarations. The function would thus appear to have parameters a, b, c, a, b, c. 

This change makes it so the function would only have parameters a, b, c. 

Note: I know there is no evidence that this claim is true, but this will be the case once testing for the AST passes is implemented. This change is needed for the development of the function-info-pass branch.

### DIFF
--- a/ASTAnalyses/ASTInfo/Specifiers/FunctionStructure/Function.py
+++ b/ASTAnalyses/ASTInfo/Specifiers/FunctionStructure/Function.py
@@ -27,7 +27,7 @@ class Function(Specifier):
         for param in self.__parameters:
             if param.get_index() == parameter.get_index():
                 return
-        
+
         self.__parameters.append(parameter)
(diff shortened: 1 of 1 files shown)


kelloggm (approved):
LGTM, and the rationale seems obvious. As usual, I'd prefer there be a test but I'll accept not having one for now.

kelloggm:
@Kevin-TD CI is failing. Why?

Kevin-TD:
Two of the tests fail because of the LLVM IR desugaring problem. It's the same tests that the main branch fails on so there's no additional error happening.

PR #46, opened 2024-08-07
Function info pass

Kevin-TD:
(To be merged after call-all-small-test-correction and test-16-logic-correction) 

AST pass + C++ class FunctionInfo created to transfer unsugared information about a function from the original code into our analysis

### DIFF
--- a/src/FunctionInfosManager.cpp
+++ b/src/FunctionInfosManager.cpp
@@ -14,7 +14,7 @@ void FunctionInfosManager::buildFunctionInfo(const TempFileManager& astInfoFile)
     logout("ran command " << readASTCommand);
 
     if (command_exit_status != 0) {
-        errs() << "error occured while building map; early exit\n";
+        errs() << "error occurred while building map; early exit\n";
         std::exit(1);
     }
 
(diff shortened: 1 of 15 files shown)


src/FunctionInfosManager.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+            // FUNCTION=_IO_vfscanf
+            auto functionChunks = rlc_util::splitString(fiChunks[0], '=');
+            if (functionChunks.size() != 2) {
+                errs() << "error 1 occurred while reading function info: bad function declaration '" << line << "'. early exit\n";
+                std::exit(1);
+            }
+            if (functionChunks[0] != "FUNCTION") {
+                errs() << "error 2 occurred while reading function info: function declaration missing \"FUNCTION\" '" << line << "'. early exit\n";
+                std::exit(1);
```
kelloggm, line 50: given that the error message includes an error code, why not exit with the corresponding code?

kelloggm (approved):
One minor comment; addressing it is up to you. Otherwise LGTM, I don't need to review again.

src/FunctionInfosManager.cpp
Kevin-TD: The error code is kind of an arbitrary number for now. Eventually, I'll make the erroring more standardized and able to produce an error code for every possible error

PR #47, opened 2024-08-08
adds testing to rlc_dataflow::variable

Kevin-TD:
This is a draft pr since I intend to add testing to all the other util functions. I just wanted to ensure that this is on the right track


kelloggm (approved):
Yes, this looks like a sensible way to test these internals.

PR #49, opened 2024-08-12
Util func testing

Kevin-TD:
Adds testing to rlc_dataflow and rlc_util functions


src/Analysis.cpp
```diff
@@ -580,11 +592,64 @@ void CodeAnalyzer::doAnalysis(Function &F, std::string optLoadFileName) {
     }
 }
 
+// assumes utilFunctionTester is actually an extended class and not UtilFunctionTester
+// (e.g., an instance of VariableTester)
```
kelloggm, line 596: You could enforce this invariant with a `dynamic_cast`, if you think it's important.

src/Analysis.cpp
```diff
@@ -580,11 +592,64 @@ void CodeAnalyzer::doAnalysis(Function &F, std::string optLoadFileName) {
...
+
+    errs() << "UTIL FUNCTION TEST PASS: " << functionName << "\n\n";
+}
+
 void CodeAnalyzer::onEnd() {
     if (BUILD_PROGRAM_LINES_BRANCH_INFO) {
         programLinesBranchesInfo.generate(cFileName, true);
     }
 
+    // rlc_dataflow function testers
```
kelloggm, line 612: This is more of a higher-level design question, but it worries me a bit that these (unit) tests are triggered by `onEnd()`. We will eventually want to run the analyzer without running the tests (e.g., to analyze a kernel module), so having the tests be triggered by the analyzer run ending is probably not the design that we want long-term. At a minimum, I think we should write a helper function (e.g., `runAllTests()` or something like that) that triggers all of the tests and call it here, so that there's only one place that needs to be changed when we eventually want to put these tests behind a flag.

src/UtilFunctionTesters/RLCUtil/HasOnlyOneBalancedParenthesesTester.cpp
```diff
@@ -0,0 +1,14 @@
...
+HasOnlyOneBalancedParenthesesTester::HasOnlyOneBalancedParenthesesTester() {}
+
+bool HasOnlyOneBalancedParenthesesTester::runTest() {
+    return rlc_util::hasOnlyOneBalancedParentheses("()") == true &&
+           rlc_util::hasOnlyOneBalancedParentheses("()()") == false &&
+           rlc_util::hasOnlyOneBalancedParentheses("(())") == false &&
+           rlc_util::hasOnlyOneBalancedParentheses("(test)") == true &&
+           rlc_util::hasOnlyOneBalancedParentheses(")()") == false &&
+           rlc_util::hasOnlyOneBalancedParentheses("(a)b") == true &&
+           rlc_util::hasOnlyOneBalancedParentheses("()b") == true;
```
kelloggm, line 13: I suggest also testing the empty string.

src/UtilFunctionTesters/RLCUtil/IsNumberTester.cpp
```diff
@@ -0,0 +1,22 @@
+#include "UtilFunctionTesters/RLCUtil/IsNumberTester.h"
+
+IsNumberTester::IsNumberTester() {}
+
+bool IsNumberTester::runTest() {
+    std::map<std::string, bool> expectedResultMap;
+    expectedResultMap["1"] = true;
+    expectedResultMap["1 "] = false;
+    expectedResultMap[""] = false;
+    expectedResultMap["1234567890123456789012345678901234567890"] = true;
```
kelloggm, line 10: We should test at least one negative number and one float (i.e., number with a decimal point in it), as well, to encode the behavior in those cases.

src/UtilFunctionTesters/RLCUtil/IsValidCVariableNameTester.cpp
```diff
@@ -0,0 +1,11 @@
+#include "UtilFunctionTesters/RLCUtil/IsValidCVariableNameTester.h"
+#include "Utils.h"
+
+IsValidCVariableNameTester::IsValidCVariableNameTester() {}
+
+bool IsValidCVariableNameTester::runTest() {
+    return rlc_util::isValidCVariableName("test") == true &&
+           rlc_util::isValidCVariableName("") == false &&
+           rlc_util::isValidCVariableName("test test") == false &&
+           rlc_util::isValidCVariableName("1test") == false;
```
kelloggm, line 10: I suggest adding the following cases:
* a valid name starting with an uppercase letter
* a valid name that ends in a number or has a number internally
* at least one test with an invalid internal character (e.g., `"f.x"`)

src/UtilFunctionTesters/RLCUtil/RemoveWhitespaceTester.cpp
```diff
@@ -0,0 +1,35 @@
+#include "UtilFunctionTesters/RLCUtil/RemoveWhitespaceTester.h"
+
+RemoveWhitespaceTester::RemoveWhitespaceTester() {}
+
+bool RemoveWhitespaceTester::runTest() {
+    std::vector<std::string> testStrings = {
+        "x",
+        "x, y, z",
+        "x,y",
+        "x, y,z"
```
kelloggm, line 10: Add a test string with leading whitespace and another with trailing whitespace.

src/UtilFunctionTesters/RLCUtil/SetToStringTester.cpp
```diff
@@ -0,0 +1,16 @@
+#include "UtilFunctionTesters/RLCUtil/SetToStringTester.h"
+
+SetToStringTester::SetToStringTester() {}
+
+bool SetToStringTester::runTest() {
+    std::set<std::string> set1 = {"test"};
+    std::set<std::string> set2 = {"test1", "test2"};
+    std::set<std::string> set3 = {""};
+    std::set<std::string> set4 = {"", ""};
```
kelloggm, line 9: Also test the empty set.

src/UtilFunctionTesters/RLCUtil/SliceStringTester.cpp
```diff
@@ -0,0 +1,10 @@
+#include "UtilFunctionTesters/RLCUtil/SliceStringTester.h"
+#include "Utils.h"
+
+SliceStringTester::SliceStringTester() {}
+
+bool SliceStringTester::runTest() {
+    return rlc_util::sliceString("str", 0, 1) == "st" &&
+           rlc_util::sliceString("str", 0, 500) == "str" &&
+           rlc_util::sliceString("str", 0, 0) == "s";
```
kelloggm, line 9: Add a test for a negative input.

src/UtilFunctionTesters/RLCUtil/SplitStringTester.cpp
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    return true;
+}
+
+bool SplitStringTester::runTest() {
+
+    return testSplitString("one two", ' ', 2, {"one", "two"}) &&
+           testSplitString("tttt", 't', 4, {"", "", "", ""}) &&
+           testSplitString("test", '\'', 1, {"test"});
```
kelloggm, line 25: Test an empty input string.

kelloggm (changes requested):
Overall LGTM

src/Analysis.cpp
Kevin-TD: I tried even passing an instance of UtilFunctionTester but that is not possible since it's an abstract class, so I'll just update the comment to be more precise; that is, updating it to say "utilFunctionTester is an extended class of UtilFunctionTester"

kelloggm (approved):
LGTM

PR #51, opened 2024-08-19
Bug fix to StructFieldToIndexMap

Kevin-TD:
Ensures empty string is not returned. This was found to be necessary during the development of un-aliasing-impl branch. Testing for StructFieldToIndexMap will be added soon to verify that this bug fix does something


kelloggm (approved):
LGTM

PR #53, opened 2024-08-22
Bug fix in DeclParser code

Kevin-TD:
Fixes bug where some struct names would simply be "struct" instead of something like "struct my_struct" or "my_struct". Bug found during development of branch test-for-struct-field-to-index-map (and the fix is proven to work in that branch as testing has been developed)


kelloggm (approved):
LGTM

PR #54, opened 2024-08-22
Adds testing for Struct Field to Index Map

Kevin-TD:


include/Debug.h
```diff
@@ -3,7 +3,7 @@
 #define DEBUG true
 #define UTIL_VARIABLE_FUNC_DEBUG false
 #define BUILD_PROGRAM_LINES_BRANCH_INFO false
-#define RUN_UTIL_FUNCTION_TESTS true
+#define RUN_UTIL_FUNCTION_TESTS false
```
kelloggm, line 6: Why is this disabled here?

kelloggm (approved):
LGTM other than the surprising fact that this change disables some other tests (maybe for debugging? I don't see a reason.).

include/Debug.h
Kevin-TD: Yeah it's just for debugging purposes. I'll re-enable the test

PR #55, opened 2024-08-22
Branch order bug fix

Kevin-TD:
Bug was found (during development of more branch lister tests) where the real branch order of one function carried on to be the branch order of another unrelated function. Patched by clearing the real branch order after running each function


kelloggm (approved):
LGTM

PR #57, opened 2024-08-23
Adds testing for Function Info + Tweaks to DeclParser logic

Kevin-TD:
Tweaks made to DeclParser are so that the testing for Function Info adheres to a certain format

### DIFF
--- a/ASTAnalyses/ASTInfo/DeclParser/DeclParser.py
+++ b/ASTAnalyses/ASTInfo/DeclParser/DeclParser.py
@@ -73,6 +73,8 @@ class DeclParser:
             return_type = return_type[:len(return_type) - 1]
 
         # if return type looks like "int *", change it to "int*"
+        # or if it looks like "int **", change it to "int**"
+        # (we assume asterisks are grouped together)
         potential_space_char_index = return_type.find("*") - 1
         if return_type[len(return_type) - 1] == "*" and return_type[potential_space_char_index] == " ":
             return_type = return_type[:potential_space_char_index] + \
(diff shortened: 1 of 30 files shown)


ASTAnalyses/ASTInfo/DeclParser/DeclParser.py
```diff
@@ -61,13 +61,23 @@ def _parse_function_decl(self, function_decl: str) -> Specifier:
...
             start_index += 1
             cur_char = function_decl[start_index]
 
+        # if return type looks like "int ", change it to "int"
+        if return_type[len(return_type) - 1] == " ":
+            return_type = return_type[:len(return_type) - 1]
+
+        # if return type looks like "int *", change it to "int*"
+        potential_space_char_index = return_type.find("*") - 1
+        if return_type[len(return_type) - 1] == "*" and return_type[potential_space_char_index] == " ":
```
kelloggm, line 77: Should this be a while loop instead? What if the return type is `"char * * "`?

kelloggm (changes requested):

ASTAnalyses/ASTInfo/DeclParser/DeclParser.py
Kevin-TD: I have yet to see the AST make the return type something like "char * *" but I have seen "char **" or "struct my_struct ****; i.e., the asterisks seem to be grouped together. So I'll add a comment noting that if it looks like "int **", change it to "int**", and that we assume asterisks are grouped together

include/Debug.h
```diff
@@ -3,7 +3,7 @@
 #define DEBUG true
 #define UTIL_VARIABLE_FUNC_DEBUG false
 #define BUILD_PROGRAM_LINES_BRANCH_INFO false
-#define RUN_UTIL_FUNCTION_TESTS true
+#define RUN_UTIL_FUNCTION_TESTS false
```
kelloggm, line 6: Should be re-enabled before merging.

PR #62, opened 2024-08-31
Adds (and adjusts) test from multi-anno-one-param-fix

Kevin-TD:
Tests moved to this branch since the multi-anno-one-param-fix branch is very outdated and will be closed


kelloggm (approved):
LGTM, but I don't really understand the rationale for the new test cases. But, there's no harm in having more...

PR #65, opened 2024-09-11
Annotation Verification for Returns Progress

Kevin-TD:


Kevin-TD:
Please verify test comments that state whether or not there should be a certain type of error

Kevin-TD:
Test comments updated @kelloggm

test/anno_verify_assignment_test1/main.c
```diff
@@ -0,0 +1,35 @@
...
+
+char* MustCall("foo2") set_mc2(char* a) {
+    return a;
+}
+
+int main() {
+    struct A a;
+    struct B b;
+
+    a.a = set_mc1(a.a); // Warning -- missing MustCall("foo1") on param 0 of set_mc1
```
kelloggm, line 27: I thought that we'd decided on Tuesday that this warning isn't necessary, because in this case we know that a.a has not been initialized (and can therefore be treated as `MustCall({})`). To be clear, I don't object to warning here: it's sound but not precise to do so.

test/anno_verify_func_test1/main.c
```diff
@@ -0,0 +1,16 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+#include "../../Annotations/Annotations.h"
+
+char* MustCall("free") foo() { // no warning
+    char* x = (char*)malloc(15);
+    char* y = (char*)malloc(15);
```
kelloggm, line 9: I'd add a comment here to note that the analysis should warn about both `x` and `y` in this function (but not for annotation verification reasons): each is leaked in the case when the other is returned. Alternatively, add `free(y)` before `return x` and vice-versa.

test/anno_verify_func_test2/main.c
```diff
@@ -0,0 +1,39 @@
...
+    if (getchar()) {
+        return x;
+    } else {
+        return x;
+    }
+}
+
+void func1(char* x) {}
+
+char* foo_causes_warning2() { // warning - missing annotation MustCall("free"), CalledMethods("func1")
```
kelloggm, line 19: I don't expect a warning about a missing `CalledMethods("func1")` annotation here, because when a caller expects an empty CalledMethods, it's okay to provide a pointer to something which has had `func1` called on it instead. More generally, for any unannotated LHS, I'd expect no CalledMethods warning - information might be lost (like the fact that `func1` really was called), but assuming nothing has been called is always safe to do.

test/anno_verify_func_test2/main.c
```diff
@@ -0,0 +1,39 @@
...
+    func1(x);
+
+    if (getchar()) {
+        return x;
+    } else {
+        return x;
+    }
+}
+
+char* Calls("func1") foo_causes_warning3() { // warning - missing annotation MustCall("free")
```
kelloggm, line 30: I agree that there should be no warning here, though, and we should keep this test case.

kelloggm: I would suggest adding another test case, though, where we do expected a warning about CalledMethods:
```
char* Calls("func1") foo_causes_warning4() {  // warning - missing annotation MustCall("free"), warning - invalid annotation Calls("func1")
    char* x = (char*)malloc(15);

    if (getchar()) {
        return x;
    } else {
        return x;
    }
}
```

test/anno_verify_assignment_test1/main.c
Kevin-TD: I'll update to comment to state it should not have a warning there

PR #67, opened 2025-07-26
Code cleanup and documentation improvement

Kevin-TD:
The following changes are summarized as follows:
- Dead code/field elimination (CFG predecessors, Annotation `targetName`, Annotation `isVerified`)
- Adding `const` and ref `&`'s where applicable (refs specifically in places where we do not need to make a copy of the argument)
- Turning `ints` into `unsigned` where applicable
- More documentation, updates to docs to be more accurate/concise, and movement of some comments to be a part of the docs to improve organization
- Typo fixes
- Name changes (changing `getName`s to `getSpecifierName`s to be more consistent with the Python end of the code; `generateStringRep` to `toString`)
- Adding backticks to code blocks or keywords

Note: updated comments on DataflowPass and related terms may not be accurate because they involve theory that I loosely remember; review those comments carefully

The code functions practically the same, and the changes made are all minor.


include/DataflowPass.h
```diff
@@ -14,11 +14,24 @@
...
+// enforces the application of transfer functions and least upper
+// bound (lub) operations, provided by from subtypes of this class)
+// only specific details on how to lub and how
+// to update a set of methods given the introduction of a new method being called.
+// this class is extended by the `MustCall` pass and `CalledMethods` pass.
 class DataflowPass {
   private:
+    // looks at the cfg to analyze the flow between
+    // the branches (nodes of the graph) and performs lub between
+    // each branch and their successors.
```
kelloggm, line 27: Here and elsewhere, it's unclear to me whether the `CFG` type represents a complete control flow graph or a single node in the logical CFG.

include/FunctionInfosManager.h
```diff
@@ -3,6 +3,12 @@
 
 #include "FunctionInfo.h"
 
+// uses the AST of a file to get information about functions *not* from the point
+// of view of the LLVM IR, but as it is seen in the code. this is important
+// as the IR can desugar functions or perform optimizations which
+// change the arity of those functions; if the arity is different that what
+// the user actually sees, then we may (or may not) analyze resources
```
kelloggm, line 10: I'm not sure what you mean by "may (or may not)" in this sentence. I understand how changing the arity of a function might complicate our analysis, but the wording here is too vague. Can you be more specific about this class' role?

include/MustCall.h
```diff
@@ -3,9 +3,14 @@
...
 #include "DataflowPass.h"
 
+// implementation of the pass that computes an overestimate
+// of the set of methods that must be called for each resource
 class MustCall : public DataflowPass {
   protected:
     void leastUpperBound(PVAliasSet &preSet, MethodsSet &curMethodsSet);
+
+    // fnName refers to corresponding deallocation/free function instead of memory
+    // allocating function
```
kelloggm, line 13: I suggest changing the name, then, because this is IMO very unintuitive.

include/ProgramRepresentation/FullFile.h
```diff
@@ -3,8 +3,26 @@
...
+
+- a `FullFile` is composed of functions (`ProgramFunction` s) that the user declares.
+- `ProgramFunction` s are composed of LLVM IR branches (`ProgramPoints` s)
+- and these branches contain information mapping each variable (and any of its aliases)
+to a set of methods called on them, which are the `DisjointPVAliasSets`.
+- these disjoint sets are composed of `PVAliasSet` s, which holds a set of program variables
+(typed `ProgramVariable`) which are must-aliases of each other, and the set of methods
+called on them (typed `MethodSet`)
+- `ProgramVariable` s are thus primitive objects that are isomorphic to the
+set of variables seen in the user's C code (or LLVM IR code – usually intermediate
```
kelloggm, line 17: I think the text in parentheses is actually correct: `ProgramVariable` refers to an LLVM IR variable. Each of these is either a variable from the C code or an intermediate variable. Please rephrase this to be more clear.

include/ProgramRepresentation/FullFile.h
```diff
@@ -3,8 +3,26 @@
...
+file. how everything in `ProgramRepresentation` is structured is as follows:
+
+- a `FullFile` is composed of functions (`ProgramFunction` s) that the user declares.
+- `ProgramFunction` s are composed of LLVM IR branches (`ProgramPoints` s)
+- and these branches contain information mapping each variable (and any of its aliases)
+to a set of methods called on them, which are the `DisjointPVAliasSets`.
+- these disjoint sets are composed of `PVAliasSet` s, which holds a set of program variables
+(typed `ProgramVariable`) which are must-aliases of each other, and the set of methods
+called on them (typed `MethodSet`)
+- `ProgramVariable` s are thus primitive objects that are isomorphic to the
```
kelloggm, line 16: What does "primitive" mean in this context? That makes me think of e.g., Java's distinction between "primitive" (unboxed) types like `int` and object types. I don't think that's the intended meaning here; these are more like the "leaves" of the structure that you're describing here.

include/ProgramRepresentation/MethodsSet.h
```diff
@@ -7,6 +7,8 @@
 // calls. all ProgramVariable's (PVs) have their own MethodsSet
 class MethodsSet {
   private:
+    // value is true if any information is passed to the internal method set but is set
+    // to and remains false if otherwise
```
kelloggm, line 11: This description is needlessly complicated. Can you phrase it instead as "true iff ..."

include/UtilFunctionTesters/UtilFunctionTester.h
```diff
@@ -5,6 +5,9 @@
 #include "Debug.h"
 #include "Utils.h"
 
+// a purely abstract class to be extended in order to represent
```
kelloggm, line 8: Is "purely abstract" different than "abstract" in this context? I think not, and "pure" has a technical meaning (deterministic and side-effect-free) when discussing programs, so avoid using the word if it's not needed.

include/UtilFunctionTesters/UtilFunctionTester.h
```diff
@@ -5,6 +5,9 @@
...
+// the fact that it is a util function that needs to have a test
+// ran
```
kelloggm, line 10: I'm not sure what "needs to have a test ran" means, or why that fact might be important.

kelloggm (changes requested):

PR #69, opened 2025-09-23
Updated CMake for version change

RobertPBlacha:
Changes CMake, uses modified version of LLVM tutor's.

### DIFF
--- a/include/RunAnalysis.h
+++ b/include/RunAnalysis.h
@@ -52,6 +52,7 @@ namespace rlc_dataflow {
 
 struct CodeAnalyzer : public PassInfoMixin<CodeAnalyzer> {
 	public:
+		//TODO: This is only left here for developmental purposes and should be removed before release
 		static bool isRequired() { return true; }
 		PreservedAnalyses run(Function &F, FunctionAnalysisManager &FAM) {
     			doAnalysis(F, F.getParent()->getSourceFileName());


CMakeLists.txt
```diff
@@ -1,18 +1,70 @@
-cmake_minimum_required(VERSION 3.10)
+# Shamelessly Ripped from llvm tutor
```
kelloggm, line 1: I presume you mean https://github.com/banach-space/llvm-tutor? When I do things like this, I usually put a permalink to the specific file/commit that I borrowed from.

CMakeLists.txt
```diff
@@ -44,6 +96,7 @@ add_llvm_library(CodeAnalyzer MODULE
   src/CalledMethods.cpp
   src/MustCall.cpp
   src/DataflowPass.cpp
+  #src/ScopeAnalyzer.cpp
```
kelloggm, line 99: Why was this file removed? It's usually a good idea to:
* keep lists of files in build scripts in sync with what's in the actual repo (so, e.g., if you want to delete something both delete the source and delete it from the script.
* generally prefer actually deleting things to commenting them out. The version control system preserves the previous state, so there's no need to worry that it'll be lost.

kelloggm (approved):
Overall looks good. Once my comments are addressed, this can be merged without another review.

CMakeLists.txt
kelloggm: (Sorry about the passive aggressive tone. Please add the comment I'm suggesting here.)

PR #74, opened 2025-12-04
Basic blocks

RobertPBlacha:
Draft for changing program representation to include basic blocks


include/ProgramRepresentation/ProgramBlock.h
```diff
@@ -0,0 +1,59 @@
...
+    // debugging function that lists variables and number of points
+    static void logoutProgramBlock(const ProgramBlock &block);
+
+    // debugging function that lists variables and number of points
+    static void logoutProgramBlock(const ProgramBlock *block);
+
+    ProgramBlock();
+    ProgramBlock(std::string blockName);
+
+    // Returns the return value associated with this program point
```
kelloggm: copy-paste error: "with this program point"

include/ProgramRepresentation/ProgramPoint.h
```diff
@@ -9,11 +9,12 @@
...
   private:
     DisjointPVAliasSets programVariableAliasSets;
-    Value *returnValue;
     std::list<ProgramPoint *> successors;
 
-    // the name is same as the branch name that shows up in the IR
-    std::string pointName;
+    // this is the first instruction num that the ProgramPoint describes
+    // To find the most current one, iterate until the next Point does not exist
+    // or the next pointName is higher than the instruction number you are
```
kelloggm: Rather than just providing these rather complicated instructions, it would be better to provide a method that does this task on behalf of a client who would otherwise need to follow these instructions.

kelloggm: (also, iterate _over what_? I think the answer is the `ProgramBlock`'s list from which you got this point...

include/ProgramRepresentation/ProgramPoint.h
```diff
@@ -9,11 +9,12 @@
...
+    int pointName;
```
kelloggm: `pointName` is an odd name for what I think is either:
* an index (e.g., into the list in the associated `ProgramBlock`, or
* a unique identifier for this point

Please add documentation explaining what these numbers _mean_. For example, I'm not sure if you intend these to be unique or not.

include/ProgramRepresentation/ProgramPoint.h
```diff
@@ -149,6 +144,12 @@ class ProgramPoint {
     argumentVar is alias information related to the pointer being assigned a new resource
     */
     void unalias(PVAliasSet* pvas, const std::string& cleanedNameOfPVToUnalias, ProgramVariable argumentVar);
+
+    /*
+     * This removes a variable from an alias set without moving it to another alias set, this is used when
+     * pointers are set to NULL instead of reassigned to other pointers
```
kelloggm, line 165: I think this has a more general use case: we should remove any variable that goes out of scope, no?

kelloggm (changes requested):
Review of documentation/header files only, not the implementations. I want to make sure I understand the purpose of each thing before I look at the code, and I'm especially confused by the `pointName` becoming an integer instead of a string.

RobertPBlacha:
@kelloggm I addressed your comments with these last two commits.

include/ProgramRepresentation/ProgramFunction.h
```diff
@@ -2,12 +2,17 @@
 #define PROGRAM_FUNCTION_H
 
 #include "ProgramRepresentation/ProgramPoint.h"
+#include "ProgramRepresentation/ProgramBlock.h"
+#include "Annotations/AnnotationHandler.h"
 
 // reflects a function from the C code. this class manages a set of points that
 // make up a function
+
```
kelloggm, line 10: This comment is now out of date and needs to be updated - a function now manages a set of blocks, not of points.

include/RunAnalysis.h
```diff
@@ -79,12 +82,16 @@ struct ResourceLeakScopeChecker : public PassInfoMixin<ResourceLeakScopeChecker>
         return true;
     }
     PreservedAnalyses run(Function &F, FunctionAnalysisManager &FAM) {
+        //PromotePass p;
+        //p.run(F, FAM);
```
kelloggm: Why was this commented out? Should it be deleted instead?

src/DataflowPass.cpp
```diff
@@ -221,32 +233,36 @@ void DataflowPass::analyzeCFG(CFG *cfg, ProgramFunction &preProgramFunction,
...
+        postProgramBlock.add(p);
 
         for(auto a : this->annotations.getAllParameterAnnotationsWithoutFields(fnName)) {
             if(a->getAnnotationType() == AnnotationType::MustCallAnnotation) {
                 int index = a->getParameterIndex();
                 auto f_iterator = this->F->args();
                 auto arg = std::next(f_iterator.begin(), index);
                 ProgramVariable var = ProgramVariable(arg);
-                auto pvas = postProgramPoint.getPVASRef(var, true);
+                auto pvas = postProgramBlock.getPoint(0)->getPVASRef(var, true);
```
kelloggm: Are we sure that program blocks are guaranteed to have at least one point? There's nothing in the program block documentation that says that we should assume that invariant.

I can imagine a situation in the future where we might want empty blocks. For example, in Java we sometimes used empty blocks to represent the implicit `finally` LUB after a `try` and a `catch`. So, we should either document the invariant that program blocks always have at least one point, or we shouldn't assume it here. I'd probably prefer the former (i.e., document the invariant), but it's up to you.

src/DataflowPass.cpp
```diff
@@ -256,32 +272,33 @@ void DataflowPass::analyzeCFG(CFG *cfg, ProgramFunction &preProgramFunction,
...
-    postProgramFunction.getProgramPointRef(priorBranch, false)->addSuccessor(postProgramFunction.getProgramPointRef(currentBranch, false));
+    postProgramFunction.getProgramBlockRef(priorBranch, false)->addSuccessor(postProgramFunction.getProgramBlockRef(currentBranch, false));
 
-    priorPostPoint->add(
-        this->programFunction.getProgramPointRef(currentBranch, true));
+    priorPostBlock->add(
+        this->programFunction.getProgramBlockRef(currentBranch, true)->getPoint(0));
 
-    if (priorPrePoint->getProgramVariableAliasSets().size() > 0) {
+    // almost certainly will cause OOB
```
kelloggm: When writing a comment like this one, it's important to say _why_, so that when we come back to this in the future we'll know what you were thinking when you wrote it.

src/DataflowPass.cpp
kelloggm: (Especially given how often you make the assumption, enforcing the invariant is almost certainly a better design.)

src/ProgramRepresentation/ProgramBlock.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "Debug.h"
+
+ProgramBlock::ProgramBlock(std::string blockName) {
+    this->blockName = blockName;
+}
+
+ProgramBlock::ProgramBlock() {}
+
+ProgramBlock::ProgramBlock(std::string pointName, ProgramBlock *prev) {
+    this->blockName = blockName;
```
kelloggm, line 16: Should this be `pointName`? `blockName` isn't in scope.

kelloggm (changes requested):

include/ProgramRepresentation/PVAliasSet.h
```diff
@@ -10,13 +10,19 @@
...
 
     // holds the methods called on the set
     MethodsSet methods;
 
   public:
     PVAliasSet();
 
+    // getters and setters for the unique id
+    int getID();
+    void setID(int newID);
```
kelloggm, line 26: Do the ids ever need to change? I would expect not, since it's rather confusing to have unique identifiers that can change.

If that's the case, it would be better not to provide an API to modify it. Instead, consider making it a parameter of the constructor instead.

include/ProgramRepresentation/PVAliasSet.h
```diff
@@ -10,13 +10,19 @@
 class PVAliasSet {
   private:
     std::list<ProgramVariable> programVariables;
+    // this is the uniquely identifying id of the PV alias set that identifies it in a function
```
kelloggm: I _think_ this means that these ids are not actually unique, only unique _within a single function_. It is a good idea for this documentation to say explicitly what the scope of the uniqueness guarantee is, because I can easily imagine this comment being misread to mean that these are _globally_ unique ids by someone in the future.

src/ProgramRepresentation/DisjointPVAliasSets.cpp
```diff
@@ -55,6 +55,16 @@ PVAliasSet *DisjointPVAliasSets::getSetRef(const std::string& cleanedName) {
     return NULL;
 }
 
+PVAliasSet *DisjointPVAliasSets::getSetRefID(int ID) {
```
kelloggm: nit: why `ID` and not `id` as the variable name? All-caps is usually only used for constants.

src/ProgramRepresentation/ProgramBlock.cpp
```diff
@@ -52,26 +52,21 @@ void ProgramBlock::update(int point) {
     ProgramPoint *one, *two;
     one = this->getPoint(point-1, false);
     two = this->getPoint(point, false);
+    if(point == 8)
```
kelloggm: Is this for debugging? I'm struggling to imagine why we might want this logic in the general case.

src/ProgramRepresentation/ProgramFunction.cpp
```diff
@@ -20,6 +20,11 @@ void ProgramFunction::addProgramBlock(ProgramBlock programBlock) {
     this->programBlocks.push_back(programBlock);
 }
 
+int ProgramFunction::getNewID() {
+    this->currentAliasNum += 1;
+    return this->currentAliasNum - 1;
```
kelloggm, line 46: nit: this implementation seems unnecessarily complicated to me. Why not `return this->currentAliasNum++;`?

kelloggm (changes requested):

src/DataflowPass.cpp
RobertPBlacha: This was changed now to work similarly to getPVASRef where an argument will specify whether to add a point if it does not already exist.

kelloggm:
LGTM at this point, so I think you can merge this once you're happy with it

src/DataflowPass.cpp
```diff
@@ -187,6 +187,11 @@ void DataflowPass::transfer(Instruction *instruction,
 
             std::string fnName = call->getCalledFunction()->getName().str();
 
+            //Previously this was handled by handeIfKnownFunction, but realloc requires changes to both argument AND result, so I broke it out into its own function
```
kelloggm, line 192: As a general rule, comments shouldn't refer to previous versions of the code: a future reader won't know what the code was like before, so it's not particularly helpful to them. A better version of this comment would read:
```
// realloc requires changes to both argument AND result, so it cannot be handled by the generic handeIfKnownFunction (which can only adjust one alias set)
```

src/MustCall.cpp
```diff
@@ -13,7 +13,19 @@ void MustCall::onAllocationFunctionCall(PVAliasSet *input,
...
+// that gets realloc'd and what it is stored into. This means that for the argument we can
+// clear the obligations and for the result we store we need to set them
+
+void MustCall::onReallocFunctionCall(PVAliasSet *result, PVAliasSet *arg, std::string &fnName) {
+    arg->clearMethods(); // Argument is being realloc'd and no longer points to valid data
+    // This might cause a problem if the analysis is expanded to have multiple
+    // different kinds of obligations (i.e. free AND close)
+    std::string free = "free";
+    this->onAllocationFunctionCall(result, free);
+    //free was hardcoded here for ease and can be changed later
```
kelloggm, line 27: Yes, this should probably be part of the specification (either via an annotation or via the list of known functions). In the Java implementation, we look up the default must-call obligations for the type (which is annotation-provided) and enforce them.

src/ProgramRepresentation/ProgramPoint.cpp
```diff
@@ -174,6 +173,21 @@ void ProgramPoint::setProgramVariableAliasSets(
...
+    if(!this->getSetID(pvas.getID())) {
+        llvm::errs() << "ADDING\n";
+        logout("> alias set (id) " << pvas.getID() << " = " << this->getSetID(pvas.getID())->toString(false, false));
+        this->addPVAS(pvas);
+        exit(1);
+    } else if(this->getSetID(pvas.getID())->getProgramVariables() == pvas.getProgramVariables()) {
+        llvm::errs() << "DUPLICATE ID " << pvas.getID() << "\n";
+        return;
+    } else {
+        exit(1);
```
kelloggm, line 185: Probably a good idea to print an error message here in addition to exiting; my impression is that this shouldn't happen, right?

include/ProgramRepresentation/DisjointPVAliasSets.h
```diff
@@ -44,20 +44,23 @@ class DisjointPVAliasSets {
     // with value val. NULL is return if it's not found.
     PVAliasSet *getSetRef(Value* val);
 
+    PVAliasSet *getSetRefID(int ID);
```
kelloggm, line 47: nit: I don't think variables should have capitalized names in C (convention is that caps are used for structs and constants/macros, but not parameters/locals/fields).

include/ProgramRepresentation/ProgramBlock.h
```diff
@@ -0,0 +1,72 @@
+#ifndef PROGRAM_BLOCK_H
+#define PROGRAM_BLOCK_H
+
+#include "ProgramRepresentation/ProgramPoint.h"
+
+// reflects a branch that holds some instructions in the IR. this class manages
```
kelloggm: Be specific in this comment that a ProgramBlock represents straightline code: that is, it never contains branches or jumps. The current comment implies this but doesn't say it outright.

include/ProgramRepresentation/ProgramBlock.h
```diff
@@ -0,0 +1,72 @@
...
+
+    // debugging function that lists variables and number of points
+    static void logoutProgramBlock(const ProgramBlock *block);
+
+    ProgramBlock();
+    //This tracks whether the block has reached fixed point
+    bool fixed;
+    ProgramBlock(std::string blockName);
+
+    // Returns the return value associated with this program block
```
kelloggm: Why does a program block have a return value associated with it? Typically we only use the phrase "return value" when talking about functions. What if this block doesn't contain any return expressions?

include/ProgramRepresentation/ProgramBlock.h
```diff
@@ -0,0 +1,72 @@
...
+    // Likely will need to be remodeled later so all resources that are owned by
+    // other parts in the program are returned
```
kelloggm: I don't understand this comment. Why?

Also nit: "refactored" rather than "remodeled"

include/ProgramRepresentation/ProgramPoint.h
```diff
@@ -2,18 +2,23 @@
...
 
 // reflects a branch that holds some instructions in the IR. this class manages
 // a point's program variables using a set of alias sets. it is effectively a
 // wrapper for DisjointPVAliasSets
 class ProgramPoint {
   private:
     DisjointPVAliasSets programVariableAliasSets;
-    Value *returnValue;
     std::list<ProgramPoint *> successors;
+    ProgramFunction *parentFunc;
```
kelloggm, line 17: The logical hierarchy here is that ProgramFunctions contain ProgramBlocks, and ProgramBlocks contain ProgramPoints. So why does a point need to know what function it is in?

kelloggm (commented):
Review of only the header files (all I had time for today); will come back for the cpp files later.

I noticed there are some unresolved comments from prior reviews as well. Please look over those and resolve them (by addressing them, or by pressing the resolve button on the ones that have already been addressed).

src/Analysis.cpp
```diff
@@ -210,11 +216,14 @@ bool onCallInst(CallInst *call, ProgramVariable receivingVar,  ProgramPoint *pro
...
 
-                logout("LHS = " << leftHandSide);
-                PVAliasSet* LHSpvas = programPoint->getPVASRef(leftHandSide, false);
+                logout("LHS IS = " << leftHandSide);
+                //PVAliasSet* LHSpvas = programPoint->getPVASRef(leftHandSide, false);
+                PVAliasSet* LHSpvas = programPoint->getPVASRef(callVar, false);
 
                 if (LHSpvas) {
-                    ret = programPoint->unalias(pvasRef, leftHandSide, call, receivingVar) || ret;
+                    llvm::errs() << "AAAA\n";
```
kelloggm: Surely this error message can be a little more helpful, even if it's supposed to be temporary...

src/CalledMethods.cpp
```diff
@@ -15,13 +15,21 @@ void CalledMethods::onDeallocationFunctionCall(PVAliasSet *input,
 void CalledMethods::onUnknownFunctionCall(PVAliasSet *input) {
     input->clearMethods();
 }
-void CalledMethods::onReallocFunctionCall(PVAliasSet *input,
-        std::string &fnName) {
-    input->clearMethods();
+
+// Because we use the mem2reg pass, there are two cases of a realloc function, the argument
+// that gets realloc'd and what it is stored into. This means that for the argument we can
+// clear the obligations and for the result we store we need to set them
```
kelloggm: This comment doesn't really align with the code, which clears both `arg` and `result`. The comment is written as if the two should be handled differently, but in the code they're handled the same. I'm not sure which is correct, but they don't match.

src/Analysis.cpp
```diff
@@ -170,420 +169,532 @@ std::vector<std::string> getAnnotationStrings(const TempFileManager& astInfoFile
...
-        logout("(load) name is " << variable(load) << " for "
-               << variable(load->getPointerOperand()));
-        std::string varName = variable(load->getPointerOperand());
+    logout("add alias for analysis loadinst");
+    return programPoint->addAlias(receivingVar, givingVar);
+}
 
-        ProgramVariable receivingVar = ProgramVariable(load);
-        ProgramVariable givingVar = ProgramVariable(load->getPointerOperand());
+// this is called as a part of the store Instruction case, otherwise no aliases are created by calls (unless we add annotations for it later?)
```
kelloggm, line 186: We will want annotations for the possibility that a call creates other aliases later, I think, but I agree that for now this is fine.

This logic is also where we will want to handle the possibility of `NonOwning` annotations on function definitions, which I think should just disable this logic at call sites.

src/Analysis.cpp
```diff
@@ -170,420 +169,532 @@ std::vector<std::string> getAnnotationStrings(const TempFileManager& astInfoFile
...
+            return change;
         }
+        ProgramVariable receivingVar = ProgramVariable(store->getOperand(1));
 
-        for (unsigned i = 0; i < call->getNumArgOperands(); ++i) {
-            ProgramVariable argumentVar = ProgramVariable(call->getArgOperand(i));
-            std::string arg = argumentVar.getCleanedName();
-            PVAliasSet *pvas = programPoint->getPVASRef(argumentVar, false);
+        // check if two structs are being aliased. the structs must refer
+        // to the same type. if they do not, they are not aliased;
```
kelloggm, line 639: Is this assumption safe in the presence of unions? I don't think so.

kelloggm: In particular, I think it's possible to use a union to create two structs that are "different" types but actually share the same memory.

src/DataflowPass.cpp
```diff
@@ -135,38 +138,78 @@ void DataflowPass::transfer(Instruction *instruction,
...
+                    pvas = newPoint->getPVASRef(assignedVar, false);
                     this->onAnnotation(pvas, returnAnno);
                 }
             }
         }
 
     } else if (CallInst *call = dyn_cast<CallInst>(instruction)) {
+
+        ProgramPoint *newPoint = inputProgramBlock->getPoint(insNum, true);
+        if(call->getName() != "") {
```
kelloggm, line 151: Can this check ever be false? I would think it's not possible for a call not to have a target, but maybe this means something like "a function pointer is being followed"? At least a comment here on why this (unintuitive) test is needed would help make this easier to understand.

src/DataflowPass.cpp
```diff
@@ -479,10 +637,8 @@ bool DataflowPass::handleIfKnownFunctionForCallInsts(CallInst *call,
 
 
     logout("call fnname = " << fnName);
-    if (this->reallocFunctions.count(fnName)) {
-        this->onReallocFunctionCall(pvas, fnName);
-        return true;
-    }
+    // the code here handling realloc functions has been moved out because the
```
kelloggm: This comment isn't useful to a future reader looking for that code unless it says where the relevant code is now.

kelloggm (approved):
I've read over all of the code. To be perfectly honest, I am not confident in it: there's enough of it and my familiarity with C is weak enough that I can't be sure that I didn't miss something. But I think this is the best you're going to get out of me in terms of a review, so once you've addressed these comments I think you can merge.

PR #75, opened 2026-04-13
Order soundness

RobertPBlacha:


include/ProgramRepresentation/ProgramVariable.h
```diff
@@ -9,6 +9,7 @@
 class ProgramVariable {
   private:
     Value *value;
+    Value *real;
```
kelloggm, line 12: A comment explaining the difference between the "value" and the "real value" would be useful here.

src/Analysis.cpp
```diff
@@ -390,7 +395,8 @@ bool onBitCastInst(BitCastInst *bitcast, ProgramPoint *programPoint) {
 
             int numFields = sourceType->getNumElements();
             for (int i = 0; i < numFields; i++) {
-                ProgramVariable sourceVar = ProgramVariable(bitcast, i);
+                //TODO: Check if this is actually needed
```
kelloggm, line 398: Is the "this" here referring to the use of the `ProgramVariable` constructor that takes a separate real value, or something else?

As a general rule for writing, it's a good idea to avoid "this" as a pronoun: its antecedent is too often ambiguous. I always try to use it as an adjective when I feel like it's appropriate. For example, if the answer to my question above is "yes", I might have written "TODO: check if calling this constructor is actually needed" or "TODO: check if passing bitcast twice is actually needed"

src/DataflowPass.cpp
```diff
@@ -325,17 +325,32 @@ void DataflowPass::analyzeCFG(CFG *cfg, ProgramFunction &preProgramFunction,
...
+                std::string name = rlc_dataflow::variable(arg) + ".addr";
+                ProgramVariable other = ProgramVariable(name);
+                postProgramBlock->getPoint(0, true)->addAlias(other, var);
                 auto pvas = postProgramBlock->getPoint(0, true)->getPVASRef(var, true);
-                this->onAnnotation(pvas, a);
+                std::cout << "OUT " << pvas->toString(false, false) << " HAS " << name << "\n\n";
+                this->onPAnnotation(pvas, a);
             }
         }
         //TODO: Add parameter annotation WITH fields
```
kelloggm, line 341: I think the new code below means you should delete this todo?

kelloggm (changes requested):
I think you should address all of the newly-introduced TODOs and make sure those changes are actually correct before merging.

